PR #4, opened 2021-04-13
added new throttle limiter for message size

leviska:


plugin/action/throttle/limiter.go
```diff
@@ -3,18 +3,20 @@ package throttle
 import (
 	"sync"
 	"time"
+
+	"github.com/ozonru/file.d/pipeline"
 )
 
 type limiter struct {
-	limit       int64 // maximum number of events per bucket
+	limit       complexLimit // maximum number of events per bucket
```
vitkovskii, line 11: let's fix the comment

plugin/action/throttle/limiter.go
```diff
@@ -55,9 +57,16 @@ func (l *limiter) isAllowed(ts time.Time) bool {
 	}
 
 	index := id - l.minID
-	l.buckets[index]++
+	switch l.limit.kind {
+	default:
```
vitkovskii, line 61: default should fatal, but empty string "" should fallthrough

plugin/action/throttle/throttle.go
```diff
@@ -58,6 +58,11 @@ type Config struct {
 	//> The default events limit that plugin allows per `interval`
 	DefaultLimit int64 `json:"default_limit" default:"5000"` //*
 
+	//> @3@4@5@6
+	//>
+	//> What we're limiting: number of messages, or total size of the messages
+	LimitType string `json:"limit_type" default:"count" options:"count|size"` //*
```
vitkovskii, line 64: in struct it named `kind`, so let's name it in all places the same

plugin/action/throttle/throttle.go
```diff
@@ -157,16 +164,21 @@ func (p *Plugin) isAllowed(event *pipeline.Event) bool {
...
 
+		// fast check with read lock
 		if !has {
-			limiter = NewLimiter(p.config.BucketInterval_, p.config.BucketsCount, rule.limit)
-			// alloc new string before adding new key to map
-			limiterKey = string(p.limiterBuff)
 			limitersMu.Lock()
-			limiters[p.pipeline][limiterKey] = limiter
+			limiter, has = limiters[p.pipeline][limiterKey]
+			// we could already write it between `limitersMu.RUnlock()` and `limitersMu.Lock()`, so we need to check again
```
vitkovskii, line 171: RLock

plugin/action/throttle/throttle_test.go
```diff
@@ -14,6 +14,53 @@ import (
...
 )
 
+type testConfig struct {
+	t           *testing.T
+	config      *Config
+	eventsTotal int
+	workTime    time.Duration
+}
+
+func (tconf *testConfig) runPipeline() {
```
vitkovskii, line 24: class variable should be named with one letter eg `func (c *testConfig) runPipeline() {`

vitkovskii (changes requested):

plugin/action/throttle/limiter.go
```diff
@@ -55,9 +58,18 @@ func (l *limiter) isAllowed(ts time.Time) bool {
...
+		logger.Fatalf("Unknown type of the limiter: \"%s\"", l.limit.kind)
```
vitkovskii, line 63: first letter should be lowercase

vitkovskii: also `%q` instead ugly `\"%s\""`

vitkovskii (changes requested):

PR #5, opened 2021-04-14
added metric for event total size

leviska:


pipeline/metrics_holder.go
```diff
@@ -53,39 +55,57 @@ func (m *metricsHolder) AddAction(metricName string, metricLabels []string) {
...
 			Subsystem:   "pipeline_" + m.pipelineName,
-			Name:        metrics.name + "_events_total",
+			Name:        metrics.name + "_events_count_total",
 			Help:        fmt.Sprintf("how many events processed by pipeline %q and #%d action", m.pipelineName, index),
 			ConstLabels: map[string]string{"gen": metricsGen},
 		}
-		counter := prometheus.NewCounterVec(opts, append([]string{"status"}, metrics.labels...))
+		cnt[0] = prometheus.NewCounterVec(opts, append([]string{"status"}, metrics.labels...))
+
+		opts.Name = metrics.name + "_events_size_total"
```
vitkovskii, line 96: it's more safe to make new opts object here

pipeline/metrics_holder.go
```diff
@@ -18,14 +18,16 @@ type metricsHolder struct {
 	registry           *prometheus.Registry
 }
 
+type counter [2]*prometheus.CounterVec
```
vitkovskii, line 21: Let's have a struct here with `count` and `size` fields instead of slice. It's confusing to choose between indices.

vitkovskii (changes requested):

PR #6, opened 2021-04-16
added journalctl input

leviska:


plugin/input/journalctl/journalctl.go
```diff
@@ -0,0 +1,84 @@
...
+
+	//> @3@4@5@6
+	//>
+	//> The filename to store offsets of processed messages.
+	OffsetsFile string `json:"offsets_file" required:"true"` //*
+
+	//> @3@4@5@6
+	//>
+	//> Additional args for journalctl
+	JournalArgs []string `json:"journal_args" default:""` //*
```
vitkovskii, line 30: what are the available options here?

plugin/input/journalctl/offset.go
```diff
@@ -0,0 +1,64 @@
...
+func (o *offsetInfo) load(r io.Reader) error {
+	if _, err := fmt.Fscanf(r, "- %d: %q", &o.offset, &o.cursor); err != nil {
+		return err
+	}
+	o.current = o.offset + 1
+	return nil
+}
+
+func (o *offsetInfo) saveFile() error {
+	file, err := os.Create(o.path)
```
vitkovskii, line 50: The file should be opened once and use the descriptor to write new information. Reopening file on each log message is bad

plugin/input/journalctl/reader.go
```diff
@@ -0,0 +1,94 @@
...
+	if config.cursor != "" {
+		res.args = append(res.args, "-c", config.cursor)
+	} else {
+		res.args = append(res.args, "-n", "all")
+	}
+	return res
+}
+
+func (r *journalReader) start() error {
+	r.config.logger.Infof("Running \"journalctl %s\"", strings.Join(r.args, " "))
```
vitkovskii, line 72: Logs should be written in lower case. Also `r.config.logger` logs plugin name by itself. Don't need to include the plugin name manually.

plugin/input/journalctl/journalctl.go
vitkovskii: Should we add `-o json ` by default? Will the plugin work without these parameters?

vitkovskii (changes requested):

plugin/input/journalctl/reader.go
leviska: need to include, because there is own config for reader: 
`
type journalReaderConfig struct {
	output   io.Writer
	cursor   string
	logger   *zap.SugaredLogger
	maxLines int
}
`

vitkovskii (changes requested):

PR #9, opened 2021-07-13
add `parse_re2` action plugin

2nick:


plugin/action/parse_re2/parse_re2.go
```diff
@@ -0,0 +1,97 @@
...
+	if len(sm) == 0 {
+		return pipeline.ActionPass
+	}
+
+	jsonNode.Suicide()
+
+	root := insaneJSON.Spawn()
+
+	fields := p.re.SubexpNames()
+	buf := bytes.NewBuffer([]byte{})
```
vitkovskii, line 79: event already has buf for general purpose data storage

https://github.com/ozonru/file.d/blob/master/pipeline/event.go#L19

vitkovskii: but you shouldn't clear it, only concat

plugin/action/parse_re2/parse_re2.go
```diff
@@ -0,0 +1,97 @@
...
+	for i := 1; i < len(fields); i++ {
+		if len(fields[i]) != 0 {
+			buf.Reset()
+			if p.config.Prefix != "" {
+				buf.WriteString(p.config.Prefix)
+			}
+			buf.WriteString(fields[i])
+
+			root.AddFieldNoAlloc(root, buf.String()).MutateToBytesCopy(root, sm[i])
```
vitkovskii, line 88: if you use `event.Buf` you can use `MutateToBytes` without copy

plugin/action/parse_re2/parse_re2.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	jsonNode.Suicide()
+
+	root := insaneJSON.Spawn()
+
+	fields := p.re.SubexpNames()
+	var bl int
+	for i := 1; i < len(fields); i++ {
+		if len(fields[i]) != 0 {
```
vitkovskii, line 80: invert if to reduce folding `if len(fields[i]) != 0 { continue }`

plugin/action/parse_re2/parse_re2.go
```diff
@@ -0,0 +1,96 @@
...
+			bl = len(event.Buf)
+			if p.config.Prefix != "" {
+				event.Buf = append(event.Buf, p.config.Prefix...)
+			}
+			event.Buf = append(event.Buf, fields[i]...)
+
+			root.AddFieldNoAlloc(root, string(event.Buf[bl:len(event.Buf)])).MutateToBytes(sm[i])
```
vitkovskii, line 87: use unsafe conversion to string

plugin/action/parse_re2/parse_re2.go
```diff
@@ -0,0 +1,96 @@
...
+	jsonNode.Suicide()
+
+	root := insaneJSON.Spawn()
+
+	fields := p.re.SubexpNames()
+	var bl int
+	for i := 1; i < len(fields); i++ {
+		if len(fields[i]) != 0 {
+			bl = len(event.Buf)
+			if p.config.Prefix != "" {
```
vitkovskii, line 82: remove if ?

PR #18, opened 2021-07-28
adding ticker to join action to flush event if no new lines appeared

2nick:


plugin/action/join/join.go
```diff
@@ -66,9 +68,11 @@ type Plugin struct {
 	controller pipeline.ActionPluginController
 	config     *Config
 
-	isJoining bool
-	initial   *pipeline.Event
-	buff      []byte
+	atomicLock int32
```
vitkovskii, line 71: use uber atomic lib here?

plugin/action/join/join.go
```diff
@@ -112,9 +116,35 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.ActionPluginP
 	p.isJoining = false
 	p.buff = make([]byte, 0, params.PipelineSettings.AvgLogSize)
 	p.logger = params.Logger
+	p.ticker = time.NewTicker(time.Second)
```
vitkovskii, line 119: make it configurable

vitkovskii (changes requested):

PR #23, opened 2021-09-03
config: add vault support and overriding by environs

Snyssfx:
- config now supports overriding with environments that begins with `FILED_` prefix, like in 12-factor apps;
- config now supports overriding by vault secrets, if you pass vault URL and token;
- add docs for configuration.


cfg/config_test.go
```diff
@@ -186,17 +190,105 @@ func TestDefaultSlice(t *testing.T) {
...
 
 func TestBase8(t *testing.T) {
 	s := &strBase8{T: "0777"}
-	err := Parse(s,nil)
+	err := Parse(s, nil)
 	assert.Nil(t, err, "shouldn't be an error")
 	assert.Equal(t, int64(511), s.T_)
 }
+
+func Test_applyEnvs(t *testing.T) {
```
vitkovskii, line 207: Lets name tests as others in this repo

cfg/config_test.go
```diff
@@ -186,17 +190,105 @@ func TestDefaultSlice(t *testing.T) {
...
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tt.wantErr)
+			}
+		})
+	}
+}
+
+type vaultMock struct{}
+
+func (v *vaultMock) GetSecret(path, key string) (string, error) {
```
vitkovskii, line 256: Maybe we should use here special variables testSecretPath=true|false, testSecretVal rather than have a special path for testing

cfg/config.go
```diff
@@ -70,11 +139,57 @@ func parseConfig(json *simplejson.Json) *Config {
...
+func tryGetSecret(vault secreter, field *simplejson.Json) (string, bool) {
+	s, err := field.String()
+	if err != nil {
+		return "", false
+	}
+	if !strings.HasPrefix(s, "vault(") || !strings.HasSuffix(s, ")") {
+		return "", false
+	}
+
+	args := strings.TrimPrefix(s, "vault(")
```
vitkovskii, line 180: Replace `vault\(` => `vault(`

PR #25, opened 2021-10-06
Event pool: add sleep instead of runtime.Gosched and therefore optimi…

Snyssfx:
…ze it on VM

### DIFF
--- a/pipeline/event.go
+++ b/pipeline/event.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"runtime"
 	"sync"
+	"time"
 
 	"github.com/ozonru/file.d/logger"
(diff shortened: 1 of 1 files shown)


pipeline/event.go
```diff
@@ -224,13 +225,20 @@ func newEventPool(capacity int) *eventPool {
...
 	x := (p.getCounter.Inc() - 1) % int64(p.capacity)
+	var i int
 	for {
 		if x < p.backCounter.Load() {
 			if p.free1[x].CAS(true, false) {
 				break
 			}
 		}
-		runtime.Gosched()
+		if i%3 == 0 {
```
vitkovskii, line 235: we need to call `runtime.Goshed()` first and then sleep

PR #37, opened 2021-12-09
Splunk output plugin fix

ansakharov:
Fixed leaks of memory in splunk output plugin.

### DIFF
--- a/plugin/output/splunk/splunk.go
+++ b/plugin/output/splunk/splunk.go
@@ -21,6 +21,8 @@ import (
 It sends events to splunk.
 }*/
 
+const outPluginType = "splunk"
+
 type Plugin struct {
 	config         *Config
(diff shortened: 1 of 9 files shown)


plugin/output/splunk/splunk.go
```diff
@@ -128,6 +131,7 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
 		root := insaneJSON.Spawn()
 		root.AddField("event").MutateToNode(event.Root.Node)
 		outBuf = root.Encode(outBuf)
+		go insaneJSON.Release(root)
```
vitkovskii, line 134: Hmm... why it's in goroutine?
Better to spawn one root before the loop and reuse it.
Don't forget to clean it like `root.Decode("{}")` on each iteration.

PR #39, opened 2021-12-20
[graceful-shutdown] Remove some races and add a bit of graceful degra…

ansakharov:
Remove some races and add some of graceful degradation scenario


cmd/file.d.go
```diff
@@ -70,7 +72,7 @@ func main() {
 
 	_, _ = maxprocs.Set(maxprocs.Logger(logger.Debugf))
 
-	go listenSignals()
+	go listenSignals(context.Background())
```
anuriq, line 75: providing this function context.Background() is redundant, it can acquire it by itself. it has only meaning if custom context is provided.

cmd/file.d.go
```diff
@@ -85,23 +87,35 @@ func start() {
 	fileD.Start()
 }
 
-func listenSignals() {
+func listenSignals(ctx context.Context) {
 	signalChan := make(chan os.Signal)
```
anuriq, line 91: this channel should be buffered

cmd/file.d.go
```diff
@@ -85,23 +87,35 @@ func start() {
...
 		case syscall.SIGHUP:
 			logger.Infof("SIGHUP received")
-			fileD.Stop()
+			fileD.Stop(ctx)
+			go func() {
+				time.Sleep(200 * time.Millisecond)
+				cancel()
+			}()
+			time.Sleep(200 * time.Millisecond)
+
```
anuriq, line 107: what happens if we sent cancel(), stop process fails, and then we say start? could there be problems?

fd/file.d.go
```diff
@@ -237,9 +238,9 @@ func (f *FileD) getStaticInfo(pipelineConfig *cfg.PipelineConfig, pluginKind pip
 	return &infoCopy, nil
 }
 
-func (f *FileD) Stop() {
+func (f *FileD) Stop(ctx context.Context) {
 	logger.Infof("stopping pipelines=%d", len(f.Pipelines))
-	_ = f.server.Shutdown(nil)
+	_ = f.server.Shutdown(ctx)
```
anuriq, line 243: maybe we can at least log the error?

pipeline/pipeline.go
```diff
@@ -366,12 +392,14 @@ func (p *Pipeline) finalize(event *Event, notifyInput bool, backEvent bool) {
 		p.totalCommitted.Inc()
 		p.totalSize.Add(int64(event.Size))
 
+		p.outSampleMu.Lock()
 		if len(p.outSample) == 0 && rand.Int()&1 == 1 {
```
anuriq, line 396: why rand thing is needed? can we leave a comment?

pipeline/pipeline.go
```diff
@@ -366,12 +392,14 @@ func (p *Pipeline) finalize(event *Event, notifyInput bool, backEvent bool) {
...
 			p.outSample = event.Root.Encode(p.outSample)
 		}
+		p.outSampleMu.Unlock()
 
-		if event.Size > p.maxSize {
-			p.maxSize = event.Size
+		if event.Size > int(p.maxSize.Load()) {
```
anuriq, line 401: is it possible to have an overflow here?

pipeline/processor.go
```diff
@@ -179,9 +179,9 @@ func (p *processor) processEvent(event *Event) (isSuccess bool, isPassed bool, e
 
 func (p *processor) doActions(event *Event) (isPassed bool) {
 	l := len(p.actions)
-	for index := event.action; index < l; index++ {
+	for index := int(event.action.Load()); index < l; index++ {
```
anuriq, line 182: maybe it's possible to avoid conversion?

pipeline/stream.go
```diff
@@ -2,6 +2,7 @@ package pipeline
 
 import (
 	"sync"
+	"sync/atomic"
```
anuriq, line 5: could this be a mistake, and uber/atomic is intended?

plugin/output/gelf/gelf.go
```diff
@@ -179,7 +180,8 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
 		p.config.BatchFlushTimeout_,
 		p.config.ReconnectInterval_,
 	)
-	p.batcher.Start()
+
+	p.batcher.Start(context.Background())
```
anuriq, line 184: context.TODO ?

plugin/output/kafka/kafka.go
```diff
@@ -106,7 +107,8 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
 		p.config.BatchFlushTimeout_,
 		0,
 	)
-	p.batcher.Start()
+
+	p.batcher.Start(context.Background())
```
anuriq, line 111: context.TODO() ?

pipeline/stream.go
ansakharov: Correct, I'll replace with uber lib.

pipeline/pipeline.go
ansakharov: I'll add comment, it's just way to say 50% possibility, like `rand.Int() % 2 == 1`

cmd/file.d.go
ansakharov: thank for notion, a fix this.

fd/file.d.go
ansakharov: We return err and throw log.Fatal above

pipeline/processor.go
ansakharov: index must be int and atomic can't be just int. I guess we can't avoid conversation.

pipeline/pipeline.go
```diff
@@ -321,9 +345,11 @@ func (p *Pipeline) In(sourceID SourceID, sourceName string, offset int64, bytes
 	event.streamName = DefaultStreamName
 	event.Size = len(bytes)
 
+	p.inSampleMu.Lock()
```
vitkovskii, line 348: We can't use lock here as it super hot path. Let's think about how to avoid mutex. Maybe cache length to atomic var and use lock only inside the condition.

pipeline/pipeline.go
vitkovskii: I think we could remove 50% possibility as it was added for debug

pipeline/pipeline.go
```diff
@@ -496,25 +531,29 @@ func (p *Pipeline) maintenance() {
 		rate := int(float64(deltaCommitted) * float64(time.Second) / float64(interval))
 		rateMb := float64(deltaSize) * float64(time.Second) / float64(interval) / 1024 / 1024
 
+		// avoid division by zero.
 		tc := totalCommitted
 		if totalCommitted == 0 {
-			tc = 1
```
vitkovskii, line 501: Why the line is removed?) It's empty condition here

vitkovskii:
let's run benchmarks for the files plugin, there's a possibility to have performance degradation

cmd/file.d.go
ansakharov: Done.

cmd/file.d.go
ansakharov: Done.

pipeline/pipeline.go
ansakharov: Done.

pipeline/pipeline.go
ansakharov: It's okay unless event msg <2GiB, which is nonsense.

pipeline/pipeline.go
ansakharov: I was in middle of refactoring. Moved it back.

plugin/output/gelf/gelf.go
ansakharov: Done.

plugin/output/kafka/kafka.go
ansakharov: Done.

cmd/file.d.go
```diff
@@ -90,18 +91,35 @@ func listenSignals() {
...
+				cancel()
+			}()
+			err := fileD.Stop(ctx)
+			if err != nil {
+				logger.Fatalf("can't stop file.d with SIGHUP: %s", err.Error())
+			}
+
 			start()
 		case syscall.SIGINT:
 			fallthrough
```
anuriq, line 111: btw this doesn't make sense. it means SIGINT is ignored

anuriq:
LGTM

fd/file.d.go
```diff
@@ -244,12 +244,15 @@ func (f *FileD) getStaticInfo(pipelineConfig *cfg.PipelineConfig, pluginKind pip
 	return &infoCopy, nil
 }
 
-func (f *FileD) Stop() {
+func (f *FileD) Stop(ctx context.Context) error {
 	logger.Infof("stopping pipelines=%d", len(f.Pipelines))
+	err := f.server.Shutdown(ctx)
 	_ = f.server.Shutdown(context.TODO())
```
ssnd, line 250: Why is server.shutdown called twice?  why context.TODO in second call?

ansakharov: Merging master gone wrong. Fixed.

cmd/file.d.go
andrewmed: fallthrough means it is processed same way as case syscall.SIGTERM

i would personally prefer though 
case syscall.SIGINT, case syscall.SIGTERM:

in one case statement because the logic is the same

andrewmed: if we want to catch SIGINT we need to add it on line 92 too

ssnd:
LGTM

PR #40, opened 2021-12-24
[mbuckets-s3] Add maintenance of multiple s3 buckets.

ansakharov:
Add maintenance of multiple buckets.


plugin/output/file/files.go
```diff
@@ -0,0 +1,44 @@
...
+// Stop stops all plugins (very useful comment).
+func (p *Plugins) Stop() {
+	for _, plug := range p.plugins {
+		plug.Stop()
+	}
+}
+
+// Exists asks if such file.Plugin exists in Plugins.
+func (p *Plugins) Exists(plugName string) (ok bool) {
+	_, ok = p.plugins[plugName]
```
anuriq, line 42: is it possible, that this map will be accessed concurrently?

plugin/output/s3/s3.go
```diff
@@ -97,96 +120,97 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
+		}
+		// sort compressed files by creation time.
+		sort.Slice(compressedFiles, p.getSortFunc(compressedFiles))
+		// upload archive.
+		for _, z := range compressedFiles {
+			splitPath := strings.Split(z, "/")
+			// multi_buckets are subdirectories of main bucket file.
+			// If it is default: /var/log/file-d.log, then bucket "s3SpecialLogs" will have route:
+			// /var/log/s3SpecialLogs/s3SpecialLogs-postfixes-with-timing.zip
+			// probableBucket will be found as ine of multi_buckets.
```
anuriq, line 189: is this a feature of minor lib? where can i read more, i didn't get this part.

plugin/output/s3/s3.go
```diff
@@ -262,3 +290,121 @@ func (p *Plugin) generateObjectName(name string) string {
...
+
+func (p *Plugin) startPlugins(Params *pipeline.OutputPluginParams, outPlugCount int) {
+	outPlugins := make(map[string]*file.Plugin, outPlugCount)
+	// Checks required bucket and exit if it's invalid.
+	exist, err := p.clients[p.config.Bucket].BucketExists(p.config.Bucket)
+	if err != nil {
+		p.logger.Panicf("could not check bucket: %s, error: %s", p.config.Bucket, err.Error())
+	}
+	if !exist {
+		p.logger.Fatalf("bucket: %s, does not exist", p.config.Bucket)
```
anuriq, line 359: why panic earlier, but fatal here?

plugin/output/s3/s3.go
```diff
@@ -262,3 +290,121 @@ func (p *Plugin) generateObjectName(name string) string {
...
+		if !exist {
+			p.logger.Fatalf("bucket from multi_backets: %s, does not exist", singleB.Bucket)
+		}
+		anyPlugin, _ := file.Factory()
+		outPlugin := anyPlugin.(*file.Plugin)
+		outPlugin.SealUpCallback = p.addFileJobWithBucket(singleB.Bucket)
+		outPlugins[singleB.Bucket] = outPlugin
+	}
+
+	p.logger.Info("clients is ready")
```
anuriq, line 381: clients are ready

plugin/output/s3/s3.go
```diff
@@ -262,3 +290,121 @@ func (p *Plugin) generateObjectName(name string) string {
...
+	p.logger.Infof("bucket: %s exists", p.config.Bucket)
```
anuriq, line 382: : is excessive

plugin/output/file/files.go
ansakharov: We concurrently read from p.plugins, but it's is, because we write only on start.

plugin/output/s3/s3.go
ansakharov: If `minio.BucketExists` returns error, it's probably network issues, so we panic -> longpanic package catch us and retrieve form dust.
But if bucket doesn't exit at s3, we are totally doomed and have no change to work correctly until bucket name in config fixed or s3 bucket created. Both actions beyond our code, we can exit with clear conscience.

plugin/output/s3/s3.go
```diff
@@ -23,6 +24,8 @@ const (
 	outPluginType      = "s3"
 	fileNameSeparator  = "_"
 	attemptIntervalMin = 1 * time.Second
+
+	bucketField = "bucket_name"
```
vano144, line 28: what do you think about putting thin in config ?

plugin/output/s3/s3.go
ansakharov: It's just little feature, so prevent intersection of other butches files.

plugin/input/http/EXAMPLE.md
```diff
@@ -0,0 +1,46 @@
+**Example:**
```
vano144, line 1: README.md?

plugin/output/file/file.go
```diff
@@ -278,3 +279,11 @@ func (p *Plugin) getStartIdx() int {
 	idx++
 	return idx
 }
+
+func (p *Plugin) SetTargetFile(name string) {
```
vano144, line 283: I don't find usage, is it necessary ?

plugin/output/file/files.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Stop stops all plugins (very useful comment).
+func (p *Plugins) Stop() {
+	for _, plug := range p.plugins {
+		plug.Stop()
+	}
+}
+
+// Exists asks if such file.Plugin exists in Plugins.
+func (p *Plugins) Exists(plugName string) (ok bool) {
```
vano144, line 41: maybe add tests on it ?

plugin/output/s3/s3.go
```diff
@@ -74,13 +82,28 @@ type Config struct {
...
+	// Event must contain `bucket_name` which value is valid s3 bucket name.
+	// Events without `bucket_name` sends to	 Bucket.
+	MultiBuckets []singleBucket `json:"multi_buckets" required:"false"`
+	Secure       bool           `json:"secure" default:"false"`
 
 	// for mock client injection
 	client *objectStoreClient
 }
 
+type singleBucket struct {
```
vano144, line 97: Since this structure is related only to the config, i guess it is better to rename it to `singleBucketConfig`, what do you think?

plugin/output/s3/s3.go
```diff
@@ -97,96 +120,97 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
 	p.logger = params.Logger
 	p.config = config.(*Config)
 
+	outPlugCount := len(p.config.MultiBuckets) + 1
```
vano144, line 123: Let's add comment about it

plugin/output/s3/s3.go
```diff
@@ -262,3 +290,122 @@ func (p *Plugin) generateObjectName(name string) string {
...
+}
+
+func (p *Plugin) setDirs(outPlugCount int) {
+	// dir for all bucket files.
+	dir, _ := filepath.Split(p.config.FileConfig.TargetFile)
+
+	targetDirs := make(map[string]string, outPlugCount)
+	targetDirs[p.config.Bucket] = dir
+	for _, singleB := range p.config.MultiBuckets {
+		targetDirs[singleB.Bucket] = fmt.Sprintf("%s%s/", dir, singleB.Bucket)
```
vano144, line 333: it is better to use filepath.Join in compressFilesInDir and here "/" will not be needed.

plugin/output/s3/s3.go
vitkovskii: yes, it'll be useful

plugin/output/s3/s3.go
```diff
@@ -97,96 +120,97 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
+		pattern := fmt.Sprintf("%s*%s", dir, p.compressor.getExtension())
+		compressedFiles, err := filepath.Glob(pattern)
+		if err != nil {
+			p.logger.Panicf("could not read dir: %s", dir)
+		}
+		// sort compressed files by creation time.
+		sort.Slice(compressedFiles, p.getSortFunc(compressedFiles))
+		// upload archive.
+		for _, z := range compressedFiles {
+			splitPath := strings.Split(z, "/")
```
vano144, line 185: It seems more readable to use filepath package.
```
dir := filepath.Dir(p)
parent := filepath.Base(dir)
```

Also here need to process case of unexpected dirs structure.

plugin/output/s3/s3.go
```diff
@@ -50,19 +53,24 @@ type Plugin struct {
...
 
-	targetDir     string
+	targetDirs    map[string]string
 	fileExtension string
-	fileName      string
+	fileNames     map[string]string
 
-	compressCh chan string
-	uploadCh   chan string
+	compressCh chan fileDto
```
vitkovskii, line 63: data transfer object? if it is make it UPPER CASE

plugin/output/s3/s3.go
```diff
@@ -74,13 +82,28 @@ type Config struct {
 	Endpoint  string `json:"endpoint" required:"true"`
 	AccessKey string `json:"access_key" required:"true"`
 	SecretKey string `json:"secret_key" required:"true"`
-	Bucket    string `json:"bucket" required:"true"`
-	Secure    bool   `json:"secure" default:"false"`
+	// Required s3 default bucket.
+	Bucket string `json:"bucket" required:"true"`
```
vitkovskii, line 86: let's rename to default_bucket

plugin/output/s3/s3.go
```diff
@@ -74,13 +82,28 @@ type Config struct {
...
+	// MultiBuckets is additional buckets, which can also receive event.
+	// Event must contain `bucket_name` which value is valid s3 bucket name.
+	// Events without `bucket_name` sends to	 Bucket.
+	MultiBuckets []singleBucket `json:"multi_buckets" required:"false"`
```
vitkovskii, line 90: We should support any bucket name that comes from field value. Hardcode here is very inconvenient for the user

plugin/output/s3/s3.go
```diff
@@ -97,96 +120,97 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
-	if !exist {
-		p.logger.Fatalf("bucket: %s, does not exist", p.config.Bucket)
-	}
-	p.logger.Info("client is ready")
-	p.logger.Infof("bucket: %s exists", p.config.Bucket)
-
-	anyPlugin, _ := file.Factory()
-	p.outPlugin = anyPlugin.(*file.Plugin)
+	// initialize minio clients.
+	p.clients = p.createMinioClients()
```
vano144, line 146: Let's return error here and panic also here, not in function

plugin/output/s3/s3.go
```diff
@@ -262,3 +290,122 @@ func (p *Plugin) generateObjectName(name string) string {
 	objectName = objectName[0 : len(objectName)-len(p.compressor.getExtension())]
 	return fmt.Sprintf("%s.%s%s", objectName, n, p.compressor.getExtension())
 }
+
+func (p *Plugin) createMinioClients() (minioClients map[string]objectStoreClient) {
+	minioClients = make(map[string]objectStoreClient)
+	// initialize minio clients object for main bucket.
+	mainClient, err := minio.New(p.config.Endpoint, p.config.AccessKey, p.config.SecretKey, p.config.Secure)
+	if err != nil || mainClient == nil {
+		p.logger.Panicf("could not create minio client, error: %s", err.Error())
```
vano144, line 299: panic will be in case of nil err

plugin/output/s3/s3.go
```diff
@@ -234,22 +258,26 @@ func (p *Plugin) uploadWork() {
...
+		p.uploadCh <- fileDto{fileName: dto.fileName, bucketName: dto.bucketName}
 	}
 }
 
-// uploadToS3 uploads compressed file to s3
-func (p *Plugin) uploadToS3(name string) error {
-	_, err := p.client.FPutObject(p.config.Bucket, p.generateObjectName(name), name, p.compressor.getObjectOptions())
+func (p *Plugin) uploadToS3(compressedDto fileDto) error {
+	_, err := p.clients[compressedDto.bucketName].FPutObject(
+		p.config.Bucket, p.generateObjectName(compressedDto.fileName),
```
vano144, line 275: Here we get client by compressedDto.bucketName, but call FPutObject with default bucket. Need to change it.

plugin/output/s3/s3.go
```diff
@@ -262,3 +290,122 @@ func (p *Plugin) generateObjectName(name string) string {
...
+	exist, err := p.clients[p.config.Bucket].BucketExists(p.config.Bucket)
+	if err != nil {
+		p.logger.Panicf("could not check bucket: %s, error: %s", p.config.Bucket, err.Error())
+	}
+	if !exist {
+		p.logger.Fatalf("bucket: %s, does not exist", p.config.Bucket)
+	}
+	p.logger.Infof("bucket %s exists", p.config.Bucket)
+
+	anyPlugin, _ := file.Factory()
```
vano144, line 363: may decrease of duplicate code here. Here is example:

```
func (p *Plugin) createPlugin(bucket string) (*file.Plugin, error) {
	exist, err := p.clients[bucket].BucketExists(bucket)
	if err != nil {
		return nil, err
	}
	if !exist {
		return nil, errors.New("todo")
	}
	p.logger.Infof("bucket %s exists", bucket)

	anyPlugin, _ := file.Factory()
	outPlugin := anyPlugin.(*file.Plugin)
	outPlugin.SealUpCallback = p.addFileJobWithBucket(bucket)
	return outPlugin, nil
}
```

plugin/output/s3/s3.go
ansakharov: I'm ok with that, `singleBucket` hasn't any attached methods it's in fact DTO.

plugin/output/s3/s3.go
ansakharov: Great idea, done.

plugin/output/file/file.go
ansakharov: Sharp eye, thx

ansakharov: Sharp eyes, it's garbage, thx.

plugin/output/s3/s3.go
ansakharov: Let's remain in field to DefaultBucket and keep json `bucket` to not break usage.

plugin/output/s3/s3.go
ansakharov: to be done

plugin/output/s3/s3.go
ansakharov: It'll help in debugging, already did it.

plugin/output/s3/s3.go
ansakharov: client can't be `nil`, I removed this condition at all.

plugin/output/s3/s3.go
ansakharov: Thx, that was a bug. I'll take up minio server probably for E2E testing.

plugin/output/s3/s3.go
ansakharov: DRY!

plugin/output/s3/s3.go
ansakharov: Thx, I added this idea, but we need need use "/"

plugin/output/s3/s3.go
ansakharov: Good point

plugin/output/s3/s3.go
ansakharov: A added both

plugin/output/file/files.go
ansakharov: I'll write couple, logic here pretty simple.

pipeline/plugin.go
```diff
@@ -119,3 +131,16 @@ const (
...
+// PlugSelector only valid value now is ByNameSelector
+// and only value is string type. If required in future it can be expanded with custom type.
+type PlugSelector struct {
+	CondType  int
+	CondValue string
+}
+
+const (
+	// UnknownSelector value is default, therefore it's safer to user it as default unknown value.
+	UnknownSelector = 0
```
vano144, line 144: better to use iota

pipeline/plugin.go
```diff
@@ -119,3 +131,16 @@ const (
 	MatchModeOr      MatchMode = 1
 	MatchModeUnknown MatchMode = 2
 )
+
+// PlugSelector only valid value now is ByNameSelector
+// and only value is string type. If required in future it can be expanded with custom type.
+type PlugSelector struct {
```
vano144, line 137: PluginSelector, better to use same prefixes for abstractions about same things

plugin/input/http/EXAMPLE.md
ansakharov: created with insane-json

plugin/output/file/files.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"sync"
+
+	"github.com/ozonru/file.d/logger"
+	"github.com/ozonru/file.d/pipeline"
+)
+
+// Plugins is an abstraction upon multiple file.Plugin, which helps reuse it.
+type Plugins struct {
+	plugins map[string]*Plugin
```
vano144, line 12: discussed the creation of a separate map for dynamic plugins

plugin/output/s3/s3.go
```diff
@@ -96,97 +211,159 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
-	pattern := fmt.Sprintf("%s*%s", p.targetDir, p.compressor.getExtension())
-	compressedFiles, err := filepath.Glob(pattern)
+// decideReceiverBucket decide which s3 bucket shall receive event.
+func (p *Plugin) decideReceiverBucket(event *pipeline.Event) string {
+	bucket := event.Root.Dig(p.config.BucketFieldInEvent).AsString()
+	// no BucketFieldInEvent in message, it's DefaultBucket, showtime.
+	if len(bucket) == 0 {
+		return p.config.DefaultBucket
+	}
+	// Bucket already exists, showtime.
```
vano144, line 277: may create plugin which alredy existed. Also be race on clients and etc.
agreed to provide a separate default client and add lock here.

plugin/output/s3/s3_internals.go
```diff
@@ -0,0 +1,175 @@
...
+		} else {
+			p.logger.Infof("can't retrieve dynamic bucket from artifacts: %s", err.Error())
+		}
+	}
+}
+
+func (p *Plugin) createOutPlugin(bucketName string) (*file.Plugin, error) {
+	exists, err := p.clients[bucketName].BucketExists(bucketName)
+	if err != nil {
+		return nil, &ErrCreateOutputPluginPanic{clientError: err.Error(), bucketName: bucketName}
```
vano144, line 114: maybe here create error using errorf ?

plugin/output/s3/s3_internals.go
```diff
@@ -0,0 +1,175 @@
...
+	p.targetDirs = targetDirs
+}
+
+func (p *Plugin) setFileNames(outPlugCount int) {
+	fileNames := make(map[string]string, outPlugCount)
+
+	_, f := filepath.Split(p.config.FileConfig.TargetFile)
+	p.fileExtension = filepath.Ext(f)
+
+	mainFileName := f[0 : len(f)-len(p.fileExtension)]
```
vano144, line 83: maybe in function and reuse it in file plugin and tests ?

plugin/output/s3/s3_internals.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	p.fileNames = fileNames
+}
+
+// Try to create buckets from dirs lying in dynamic_dirs route
+func (p *Plugin) createPlugsFromDynamicBucketArtifacts() {
+	dynamicDirsPath := filepath.Join(p.targetDirs[p.config.DefaultBucket], dynamicBucketDir)
+	dynamicDir, err := ioutil.ReadDir(dynamicDirsPath)
+	if err != nil {
+		p.logger.Info("can't access dynamic dir: %s", err.Error())
```
vano144, line 96: maybe return here and log with Error level here?

ansakharov: Return missed, but log level should be Info: absence of dynamic dir is okay, it's basic scenario.

plugin/output/s3/s3.go
ansakharov: After a while I went to this:
```go
func (p *Plugin) decideReceiverBucket(event *pipeline.Event) string {
	bucketName := event.Root.Dig(p.config.BucketFieldInEvent).AsString()
	// No race, because p.config.DefaultBucket immutable, RO.
	if len(bucketName) == 0 {
		return p.config.DefaultBucket
	}

	// Concurrent safety lays on p.outPlugins itself.
	if p.outPlugins.Exists(bucketName) {
		return bucketName
	}

	// Lock for creating new dynamic bucket
	p.dynamicPlugCreationMu.Lock()

	// Some one may already create bucket.
	if p.outPlugins.IsDynamic(bucketName) {
		return bucketName
	}
	// If not we will create!
	if created := p.tryRunNewPlugin(bucketName); created {
		// Succeed, return new bucketName.
		return bucketName
	}
	p.dynamicPlugCreationMu.Unlock()

	// Failed to create, fallback on DefaultBucket.
	return p.config.DefaultBucket
}
```

plugin/output/s3/s3_internals.go
ansakharov: How you will switch between `ErrCreateOutputPluginPanic` and `ErrCreateOutputPluginNoSuchBucket` after they become just wrapped errors like: `return nil, fmt.Errorf("could not check bucket %s, error: %w", bucketName, err)` and `fmt.Errorf("bucket: %s, does not exist", err.bucketName)`? They are generic errors, error.Is will not pass.

vano144:
```
err := errors.New("test error")
wrappedErr := fmt.Errorf("could not check bucket %s, error: %w", "test bucket", err)
fmt.Println(errors.Is(wrappedErr, err))
```

it returns true

ansakharov: It works fine in same function, which isn't. We return this error to upper level at `Start` function, where we only decide to exit or panic (because it's more obvious, than die in depths).

pipeline/plugin.go
```diff
@@ -33,6 +33,18 @@ type OutputPlugin interface {
 	Out(*Event)
 }
 
+type PluginsStarterData struct {
+	Config AnyConfig
+	Params *OutputPluginParams
+}
+type PluginsStarterMap map[string]PluginsStarterData
+
+type OutputPlugins interface {
```
vano144, line 42: I don't find usage, is it necessary ?

plugin/output/s3/s3.go
```diff
@@ -96,97 +217,166 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
+	// Bucket exists.
+	if p.outPlugins.Exists(bucketName) {
+		return bucketName
+	}
+
+	// To avoid concurrent creation of bucketName plugin add take lock.
+	p.dynamicPlugCreationMu.Lock()
+	// Probably other worker created plugin concurrently, need check dynamic bucket one more time.
+	if p.outPlugins.IsDynamic(bucketName) {
+		return bucketName
```
vano144, line 293: Unlock is missing here and below in next return, maybe add test on it ?

plugin/output/s3/s3_internals.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	outPlugins[p.config.DefaultBucket] = outPlugin
+	p.logger.Infof("bucket %s exists", p.config.DefaultBucket)
+
+	p.createPlugsFromDynamicBucketArtifacts(targetDirs)
+	// If multi_buckets described on file.d config, check each of them as well.
+	for _, singleBucket := range p.config.MultiBuckets {
+		outPlugin, err := p.createOutPlugin(singleBucket.Bucket)
+		if err != nil {
+			return err
```
vano144, line 141: Do we need to stop other already created plugins in case of err?

plugin/output/file/files.go
```diff
@@ -0,0 +1,82 @@
...
+
+func NewFilePlugins(plugins map[string]*Plugin) *Plugins {
+	return &Plugins{plugins: plugins}
+}
+
+func (p *Plugins) Out(event *pipeline.Event, selector pipeline.PluginSelector) {
+	switch selector.CondType {
+	case pipeline.ByNameSelector:
+		p.mu.RLock()
+		defer p.mu.RUnlock()
```
leviska, line 28: этот дефер сработает при выходи из функции, а не из выхода из switch case
сейчас оно работает, но может запутать в будущем
как вариант можно сделать
```
case ...:
  func() {
    p.mu.RLock()
    defer p.mu.RUnlock()
    
    p.plugins...
  }
```

plugin/output/file/files.go
```diff
@@ -0,0 +1,82 @@
...
+		plug.Stop()
+	}
+}
+
+// Add new plugin to plugs.
+func (p *Plugins) Add(plugName string, plug *Plugin) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.plugins[plugName] = plug
```
leviska, line 61: тут не dynamicPlugins нужен? если нет, то он вроде нигде не используется

plugin/output/file/file_test.go
```diff
@@ -66,7 +66,7 @@ func TestGetStartIdx(t *testing.T) {
 			fileName:      file[0 : len(file)-len(extension)],
 		}
 
-		// create files
+		// create plugins
```
leviska, line 69: а где здесь плагины?

pipeline/plugin.go
ansakharov: Correct, I'll add interface dep in s3 plugin or drop this iface.

plugin/output/s3/s3.go
ansakharov: Tests will come soon

plugin/output/file/files.go
ansakharov: I assume, that RLock usage is pretty clear there and new conditions won't add complexity:
```go
switch selector.CondType {
	case pipeline.ByNameSelector:
		p.mu.RLock()
		defer p.mu.RUnlock()

		p.plugins[selector.CondValue].Out(event)
        case pipeline.BySecondSelector:
               pluginName := someFunc(selector.CondValue)

                p.mu.RLock()
		defer p.mu.RUnlock()

		p.plugins[pluginName].Out(event)
	case pipeline.UnknownSelector:
		logger.Fatalf("PluginSelector type didn't set for event: %v, selector: %v", event, selector)
	}
```
If it's complicated I can add func, like you suggested, but it seems too much.

plugin/output/file/files.go
ansakharov: You totally right, thx for pointing out at this bug. I'll write tests specially for dynamic plugins now.

plugin/output/file/file_test.go
ansakharov: IDE played me out.

plugin/output/s3/s3_internals.go
ansakharov: It'll go up to main func and die there. I think we are ok.

plugin/output/file/files.go
leviska: problem not in the switch, but the code afterwards
it's easy to do something like 
```go
func (p *Plugins) Out(event *pipeline.Event, selector pipeline.PluginSelector) {
	switch selector.CondType {
	case pipeline.ByNameSelector:
		p.mu.RLock()
		defer p.mu.RUnlock()

		p.plugin...
	}
	doSomeOtherStuff()
}
```

and it will not be so obvious, that `doSomeOtherStuff` still in lock

plugin/output/s3/s3.go
```diff
@@ -96,97 +217,167 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
+		return p.config.DefaultBucket
+	}
+	// Bucket exists.
+	if p.outPlugins.Exists(bucketName) {
+		return bucketName
+	}
+
+	// To avoid concurrent creation of bucketName plugin add take lock.
+	p.dynamicPlugCreationMu.Lock()
+	defer p.dynamicPlugCreationMu.Unlock()
```
leviska, line 291: may be move this locks and the check afterwards inside `p.tryRunNewPlugin(bucketName)`, so the lock will be inside and nobody will forget about taking it

plugin/output/s3/mock/s3.go
```diff
@@ -0,0 +1,142 @@
+// Code generated by MockGen. DO NOT EDIT.
```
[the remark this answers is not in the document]
ansakharov: I will add it!

plugin/output/s3/s3.go
```diff
@@ -96,97 +217,167 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
 	}
-	p.client = minioClient
+	p.defaultClient = defaultClient
+	p.clients = clients
+
+	err = p.startPlugins(params, outPlugCount, targetDirs, fileNames)
+	switch err.(type) {
+	case nil:
+		break
+	case *ErrCreateOutputPluginPanic:
```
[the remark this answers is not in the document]
ansakharov: Yeah, @vano144 mentioned same thing.

plugin/output/s3/s3.go
```diff
@@ -19,10 +23,85 @@ import (
...
+          secret_key: "secret_key3"
+          bucket: "bucket-logs-3"
+```
+}*/
+
 const (
 	outPluginType      = "s3"
 	fileNameSeparator  = "_"
 	attemptIntervalMin = 1 * time.Second
+	dirSep             = "/"
```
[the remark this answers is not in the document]
ansakharov: Во-первых это красиво)

plugin/output/s3/s3.go
ansakharov: Good point.

plugin/output/s3/s3.go
```diff
@@ -96,97 +217,167 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
-		p.logger.Fatalf("bucket: %s, does not exist", p.config.Bucket)
+	if !exists {
+		p.logger.Infof("Bucket %s doesn't exist in s3, message sent to DefaultBucket", bucketName)
+		return false
 	}
-	p.logger.Info("client is ready")
-	p.logger.Infof("bucket: %s exists", p.config.Bucket)
 
+	dir, _ := filepath.Split(p.config.FileConfig.TargetFile)
+	bucketDir := filepath.Join(dir, dynamicBucketDir, bucketName) + dirSep
```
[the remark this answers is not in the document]
ansakharov: Yeah, it must be here. I added it on purpose, it wasn't obvious do we need it at all: `+ "/"`

plugin/output/s3/s3.go
```diff
@@ -96,97 +217,167 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
+		// Succeed, return new bucketName.
+		return bucketName
+	}
+
+	// Failed to create, fallback on DefaultBucket.
+	return p.config.DefaultBucket
+}
 
-	exist, err := p.client.BucketExists(p.config.Bucket)
+func (p *Plugin) tryRunNewPlugin(bucketName string) (isCreated bool) {
```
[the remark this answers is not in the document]
ansakharov: Already logging it on upper level:
```go
                 created := p.tryRunNewPlugin(f.Name())
		if created {
			p.logger.Infof("dynamic bucket '%s' retrieved from artifacts", f.Name())
		} else {
			p.logger.Infof("can't retrieve dynamic bucket from artifacts: %s", err.Error())
		}
```

It's not an error, that need some actions on upper level, so log it right here and fallback to default plugin in failsafe way.

plugin/output/file/files.go
ansakharov: Added

plugin/output/s3/mock/s3.go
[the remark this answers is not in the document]
ansakharov: Added.

Makefile
```diff
@@ -63,3 +69,7 @@ push-images-all: push-images-version push-images-latest
 lint:
 	# installation: https://golangci-lint.run/usage/install/#local-installation
 	golangci-lint run --new-from-rev=${UPSTREAM_BRANCH}
+
+.PHONY: mock
+mock:
+	go generate mock_gen.go
```
vano144, line 75: maybe add one more line and delete mock_gen.go ?
```
go get github.com/golang/mock/mockgen
mockgen -source=plugin/output/s3/s3.go -destination=plugin/output/s3/mock/s3.go
```

ansakharov: I tried both this method and declaring mockgen in Makefile, both are fine.
Seems fine, is it really an issue?

cmd/file.d_test.go
```diff
@@ -150,3 +152,61 @@ func runWriter(tempDir string, files int) {
...
+/*
+Plugins registered automatically after importing by init() function:
+	_ "github.com/ozontech/file.d/plugin/output/devnull"
+	_ "github.com/ozontech/file.d/plugin/output/elasticsearch"
+Moving plugin in sub dir in plugin dir will quit registration quietly.
+To prevent this let's check that DefaultPluginRegistry contains all plugins.
+Plugins "dmesg", "journalctl" linux based, they contain tag: //go:build linux.
+We don't check them.
+*/
+func TestThatPluginsAreImported(t *testing.T) {
```
vitkovskii, line 165: А зачем тестить? Если тут хардкод и в cmd тоже хардкод? Лучше сделать сканирование папок плагинов и потом проверить, что все они есть в cmd. Можно рассмотреть такой скрипт в мейк файле даже

pipeline/plugin.go
```diff
@@ -119,3 +125,18 @@ const (
...
+// and only value is string type. If required in future it can be expanded with custom type.
+type PluginSelector struct {
+	CondType  ConditionType
+	CondValue string
+}
+
+type ConditionType int
+
+const (
+	// UnknownSelector value is default, therefore it's safer to user it as default unknown value.
```
vitkovskii, line 139: to use

pipeline/plugin.go
```diff
@@ -119,3 +125,18 @@ const (
 	MatchModeOr      MatchMode = 1
 	MatchModeUnknown MatchMode = 2
 )
+
+// PluginSelector only valid value now is ByNameSelector
```
vitkovskii, line 129: the only valid

vitkovskii: for now

pipeline/plugin.go
```diff
@@ -119,3 +125,18 @@ const (
...
+// and only value is string type. If required in future it can be expanded with custom type.
```
vitkovskii, line 130: it can be expanded with a custom type in the future

plugin/output/file/file.go
```diff
@@ -203,7 +210,7 @@ func (p *Plugin) write(data []byte) {
 
 func (p *Plugin) createNew() {
 	p.tsFileName = fmt.Sprintf("%d%s%s%s", time.Now().Unix(), fileNameSeparator, p.fileName, p.fileExtension)
-	logger.Errorf("tsFileName in createNew=%s", p.tsFileName)
+	logger.Infof("tsFileName in createNew=%s", p.tsFileName)
```
vitkovskii, line 213: это вроде как для дебага, нужно удалить?

plugin/output/file/file_test.go
```diff
@@ -76,7 +76,7 @@ func TestGetStartIdx(t *testing.T) {
 		idx := p.getStartIdx()
 		assert.EqualValues(t, tc.expectedIdx, idx)
 
-		//	close files
+		//	close files.
```
vitkovskii, line 79: а чего тут таб?

plugin/output/file/files.go
```diff
@@ -0,0 +1,93 @@
...
+func NewFilePlugins(plugins map[string]PluginInterface) *Plugins {
+	return &Plugins{plugins: plugins, dynamicPlugins: make(map[string]PluginInterface)}
+}
+
+func (p *Plugins) Out(event *pipeline.Event, selector pipeline.PluginSelector) {
+	switch selector.CondType {
+	case pipeline.ByNameSelector:
+		p.out(event, selector)
+	default:
+		logger.Fatalf("PluginSelector type didn't set for event: %v, selector: %v", event, selector)
```
vitkovskii, line 29: event=, selector=

plugin/output/file/files_test.go
```diff
@@ -0,0 +1,137 @@
+package file
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPlugins_Exists(t *testing.T) {
```
vitkovskii, line 9: используй стиль именования тестов как в других местах

plugin/output/s3/s3.go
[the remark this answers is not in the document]
vitkovskii: может path.Join использовать? кстати да, вместо Sprintf и все такое

plugin/output/s3/s3.go
```diff
@@ -47,38 +127,91 @@ type compressor interface {
...
+	//> Event must contain `bucket_name` field which value is s3 bucket name.
+	//> Events without `bucket_name` sends to DefaultBucket.
+	MultiBuckets []singleBucketConfig `json:"multi_buckets" required:"false"` //*
+	//> @3@4@5@6
+	//> s3 connection secure option.
+	Secure bool `json:"secure" default:"false"` //*
+	//> @3@4@5@6
+	//> BucketFieldInEvent field change destination bucket of event to fields value.
+	// Fallback to DefaultBucket if BucketFieldInEvent bucket doesn't exist.
+	BucketFieldInEvent string `json:"bucket_field_in_event" default:"bucket_name"` //*
```
vitkovskii, line 190: давай просто BucketEventField и дефолтно пустое

plugin/output/file/file.go
ansakharov: It's rather useful info, let leave it.

plugin/output/s3/s3.go
```diff
@@ -47,38 +127,91 @@ type compressor interface {
...
+	for _, bucket := range c.MultiBuckets {
+		if bucketName == bucket.Bucket {
+			return true
+		}
+	}
+
+	return false
+}
+
+type singleBucketConfig struct {
```
vitkovskii, line 207: давай это объявим выше места, где это используется

vitkovskii: А эту штуку нельзя тупо заэмбедидь в Config, а то дублирование какое-то получается

pipeline/plugin.go
ansakharov: My upper intermediate more rather intermediate :(

plugin/output/s3/s3.go
```diff
@@ -93,100 +226,176 @@ func Factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
 func (p *Plugin) Stop() {
-	p.outPlugin.Stop()
+	p.outPlugins.Stop()
 }
 
 func (p *Plugin) Out(event *pipeline.Event) {
-	p.outPlugin.Out(event)
+	p.outPlugins.Out(event, pipeline.PluginSelector{
+		CondType:  pipeline.ByNameSelector,
+		CondValue: p.decideReceiverBucket(event),
```
vitkovskii, line 289: getReceiverBucket

Makefile
ansakharov: Done

plugin/output/s3/s3.go
vitkovskii: getBucketName еще лучше

plugin/output/s3/s3.go
```diff
@@ -234,22 +443,33 @@ func (p *Plugin) uploadWork() {
 
 // compressWork compress file from channel and then delete source file
 func (p *Plugin) compressWork() {
-	for f := range p.compressCh {
-		compressedName := p.compressor.getName(f)
-		p.compressor.compress(compressedName, f)
+	for DTO := range p.compressCh {
```
vitkovskii, line 446: dto

plugin/output/s3/s3_test.go
```diff
@@ -47,23 +52,56 @@ func (m mockClient) FPutObject(bucketName, objectName, filePath string, opts min
...
 
+type putWithErr struct {
+	ctx    context.Context
+	cancel context.CancelFunc
+}
+
+func (put *putWithErr) fPutObjectErr(bucketName, objectName, filePath string, opts minio.PutObjectOptions) (n int64, err error) {
+	select {
+	case <-put.ctx.Done():
+		fmt.Println("put object")
```
vitkovskii, line 76: а куда logger.* делся?)))

plugin/output/s3/s3_test.go
```diff
@@ -161,7 +208,194 @@ func TestStart(t *testing.T) {
...
+		assert.Equal(t, 1, len(match))
+		test.CheckZero(t, match[0], "log file is not empty")
+	}
+
+	size2 := test.CheckNotZero(t, fileName, "s3 data missed after second pack")
+	assert.True(t, size2 > size1)
+
+	// failed during writing
+	test.SendPack(t, p, tests.thirdPack)
+	time.Sleep(200 * time.Millisecond)
```
vitkovskii, line 378: Использовать слипы в тестах — дурной тон. Давай попробуем избавиться. В других тестах, конечно они есть, но там тоже нужно будет их удалить в какой-то момент.

plugin/output/s3/s3.go
ansakharov: Move upper and embedded

cmd/file.d_test.go
ansakharov: Such script won't save us if I move plugin files in subdir.
Before:
```
plugins
  |--> action
           | --> add_host
                     |--> add_host.go
```
After:
```
plugins
  |--> action
            |--> add_host
                      | --> new_sub_dir
                               | --> add_host.go
```

Import  `_ "github.com/ozontech/file.d/plugin/action/add_host"` will work, but `init()` from `add_host.go` won't be called.
This test prevents from breaking old plugins, but doesn't helps with new one.

plugin/output/file/file.go
```diff
@@ -19,6 +19,12 @@ import (
 	"golang.org/x/net/context"
 )
 
+type PluginInterface interface {
```
vitkovskii, line 22: Кстати название интерфейса не должено содержать слово Interface ))) Plugable наверное нужно написать

plugin/output/file/file.go
```diff
@@ -53,6 +59,7 @@ const (
 
 type Config struct {
 	//> File name for log file.
+	// 	defaultTargetFileName = TargetFile default value
```
vitkovskii, line 62: Таб

plugin/output/file/files_test.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, tCase := range cases {
+		t.Run(tCase.name, func(t *testing.T) {
+			plugins := tCase.prepareFunc()
+			exists := plugins.Exists(tCase.search)
+			require.Equal(t, tCase.expected, exists)
+		})
+	}
+}
+
+func TestPlugins_IsStatic(t *testing.T) {
```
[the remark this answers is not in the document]
vitkovskii: ???????

plugin/output/file/file.go
ansakharov: okay :)

plugin/output/s3/s3_test.go
ansakharov: Sleep already in this tests every where, nothing new added here, to cut it I need add sync control mechanism. I'll fix this, but not in this PR.

plugin/output/s3/s3.go
```diff
@@ -93,100 +226,176 @@ func Factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
+		pattern := fmt.Sprintf("%s*%s", dir, p.compressor.getExtension())
+		compressedFiles, err := filepath.Glob(pattern)
+		if err != nil {
+			p.logger.Panicf("could not read dir: %s", dir)
+		}
+		// sort compressed files by creation time.
+		sort.Slice(compressedFiles, p.getSortFunc(compressedFiles))
+		// upload archive.
+		for _, z := range compressedFiles {
+			splitPath := strings.Split(z, "/")
```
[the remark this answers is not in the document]
ansakharov: It was redundant, removed it at all.

PR #44, opened 2022-01-24
[fix-filewatcher]

ansakharov:


Makefile
```diff
@@ -11,10 +11,16 @@ deps:
...
-	go test -coverprofile=coverage.out ./...
+	go test -short -coverprofile=coverage.out ./...
 	go tool cover -html=coverage.out
 	rm coverage.out
 
+.PHONY: test-short
+test-short:
+	go test ./fd/ -v -count 1 -short
+	go test ./pipeline/ -v -count 1 -short
+	go test ./plugin/... -v -count 1 -short
```
anuriq, line 22: so we need to manage this list here? what if someone forgets to add here?

pipeline/pipeline.go
```diff
@@ -264,7 +264,7 @@ func (p *Pipeline) GetOutput() OutputPlugin {
 func (p *Pipeline) In(sourceID SourceID, sourceName string, offset int64, bytes []byte, isNewSource bool) uint64 {
 	length := len(bytes)
 
-	// don't process shit
+	// don't process shit.
```
anuriq, line 267: if we edit this, let's write something more thoughtful

plugin/input/file/worker.go
```diff
@@ -83,18 +85,23 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
 					break
 				}
 
+				// \n is line separator, -1 means that file doesn't have new valid logs.
```
anuriq, line 88: what about the last line in a file?

plugin/input/journalctl/reader.go
```diff
@@ -59,6 +62,7 @@ func readLines(r io.Reader, config *journalReaderConfig) {
 	}
 }
 
+//nolint:deadcode,unused
```
anuriq, line 65: if it's dead code, let's remove it

anuriq (changes requested):
please, add meaningful commit messages

plugin/input/file/worker.go
```diff
@@ -54,7 +55,9 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
 			logger.Fatalf("can't get offset, file %d:%s seek error: %s", sourceID, sourceName, err.Error())
 		}
 
+		// did reach end of file.
```
vitkovskii, line 58: have end of file reached

plugin/input/file/worker.go
```diff
@@ -54,7 +55,9 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
...
 		isEOF := false
+		// did put any event.
```
vitkovskii, line 60: have any event put

plugin/input/file/worker.go
```diff
@@ -63,18 +66,17 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
 
 		accumBuffer = accumBuffer[:0]
 		for {
+			// on each iteration cup is readBufferSize, so cut all extra bytes.
```
vitkovskii, line 69: cup => cap?

plugin/input/file/worker.go
```diff
@@ -63,18 +66,17 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
...
-
+			// if we read to end of file time to check truncation etc and process next job.
 			if err == io.EOF || read == 0 {
 				isEOF = true
 				break
-			}
-
-			if err != nil {
-				logger.Fatalf("file %d:%s read error, %s read=%d", sourceID, sourceName, read, err.Error())
+			} else if err != nil {
```
vitkovskii, line 78: if think without else if the code is more clean

plugin/input/file/worker.go
```diff
@@ -138,21 +145,35 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
...
+	}
+
+	// files truncated from time to time, after logs from file was processed.
+	// Position > stat.Size() means that data was truncated and
+	// caret pointer must be moved to start of file.
+	if totalOffset > stat.Size() {
+		jobProvider.truncateJob(job)
+	}
+
+	// finish job. Until any new writes to this file it doesn't bother us.
```
vitkovskii, line 175: Mark job as done till new lines has appeared

plugin/input/journalctl/reader.go
ansakharov: journalctl linux only, that's for golangci on mac/win

Makefile
ansakharov: it's same as `make test` but with `-short` flag.

plugin/input/file/worker.go
ansakharov: this works fine, we don't write empty lines.

plugin/input/file/worker.go
```diff
@@ -138,21 +147,35 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
 			}
 		}
 
-		// check if file was truncated
+		totalOffset := lastOffset + readTotal
```
vano144, line 150: Maybe it should be calculated on 153 ?

plugin/input/file/worker.go
```diff
@@ -138,21 +147,35 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
...
+		// check if file was truncated.
 		if isEOF {
-			stat, err := file.Stat()
+			err := w.processEOF(file, job, jobProvider, totalOffset)
```
vano144, line 153: It can be replaced by with one line:

```
if err := w.processEOF(file, job, jobProvider, totalOffset); err != nil {
```

ansakharov: done

vano144:
LGTM

plugin/input/file/worker.go
ansakharov: done

plugin/input/file/watcher.go
```diff
@@ -56,12 +56,11 @@ func (w *watcher) start() {
 	eventsCh := make(chan notify.EventInfo, 128)
 	w.watcherCh = eventsCh
 
-	events := []notify.Event{notify.Create}
+	events := []notify.Event{notify.Create, notify.Rename}
```
vano144, line 59: What about notify.Remove event ?

ansakharov: Added.

PR #46, opened 2022-01-31
Add optional metric to mask plugin

ansakharov:


plugin/README.md
```diff
@@ -53,6 +53,54 @@ So you can use Elasticsearch filebeat output plugin to send data to `file.d`.
...
+```
+# run server.
+# config.yaml should contains yaml config above.
+go run cmd/file.d.go --config=config.yaml
+
+# now do requests.
+curl "localhost:9200/_bulk" -H 'Content-Type: application/json' -d \
+'{"index":{"_index":"index-main","_type":"span"}}
+{"message": "hello", "kind": "normal"}
+
```
vano144, line 100: missing '

plugin/action/mask/mask.go
```diff
@@ -1,6 +1,8 @@
 package mask
 
 import (
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
```
vano144, line 5: need to sort imports here

plugin/action/mask/mask.go
```diff
@@ -25,27 +27,39 @@ pipelines:
       - mask:
         re: "\b(\d{1,4})\D?(\d{1,4})\D?(\d{1,4})\D?(\d{1,4})\b"
         groups: [1,2,3]
+        metric_subsystem_name: "some_name"
```
vano144, line 30: it is located on the same level with masks, need to fix examples

```
masks:
      - mask:
        re: "\b(\d{1,4})\D?(\d{1,4})\D?(\d{1,4})\D?(\d{1,4})\b"
        groups: [1,2,3]
metric_subsystem_name: "some_name"
```

plugin/action/mask/mask.go
```diff
@@ -204,19 +234,27 @@ func getValueNodeList(currentNode *insaneJSON.Node, valueNodes []*insaneJSON.Nod
...
 			p.sourceBuf = p.maskBuf
+			if locApplied {
+				maskApplied = true
+			}
 		}
 		v.MutateToString(string(p.maskBuf))
 	}
+	if p.logMaskAppeared.Load() && maskApplied {
+		p.counterMetric.Inc()
+		p.logger.Infof("mask appeared to event, output string: %s", string(p.maskBuf))
```
vano144, line 256: p.maskBuf changes on each iteration of loop you will see here only one value from event

plugin/action/mask/mask.go
```diff
@@ -25,27 +27,39 @@ pipelines:
...
 type Plugin struct {
-	config     *Config
-	sourceBuf  []byte
-	maskBuf    []byte
-	valueNodes []*insaneJSON.Node
-	logger     *zap.SugaredLogger
+	config          *Config
+	sourceBuf       []byte
+	maskBuf         []byte
+	logMaskAppeared atomic.Bool
```
vano144, line 46: bool? you only change this from constructor

ansakharov: Bool'll cause race in case of panic during start of file.d and concurrent second `Start` call after recovering.

plugin/action/mask/mask_test.go
```diff
@@ -22,75 +22,85 @@ const (
...
-			input:    []byte("12.34.5678"),
-			masks:    Mask{Re: `\d`, Groups: []int{0}},
-			expected: []byte("**.**.****"),
-			comment:  "all digits should be masked",
+			name:         "simple test",
+			input:        []byte("12.34.5678"),
+			masks:        Mask{Re: `\d`, Groups: []int{0}},
+			expected:     []byte("**.**.****"),
+			comment:      "all digits should be masked",
+			mustBeMusked: true,
```
anuriq, line 38: Musked -> Masked ?

anuriq (changes requested):

plugin/action/mask/mask_test.go
ansakharov: Elon disapproves to this decision.

anuriq:
LGTM

ansakharov:
OK from @vano144, @anuriq.

PR #48, opened 2022-02-11
some minor improvements (no logic change)

anuriq:
- remove unused functions and variables
- put pointers in sync.Pool for http input plugin (better performance)

### DIFF
--- a/offset/offset.go
+++ b/offset/offset.go
@@ -30,10 +30,7 @@ func (o *Offset) Load() error {
 	}
 	defer file.Close()
 
-	if err := o.Callback.Load(file); err != nil {
-		return err
-	}
-	return nil
+	return o.Callback.Load(file)
 }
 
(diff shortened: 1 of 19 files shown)


pipeline/pipeline.go
```diff
@@ -273,7 +273,7 @@ func (p *Pipeline) In(sourceID SourceID, sourceName string, offset int64, bytes
 	}
 
 	event := p.eventPool.get()
-	dec := decoder.NO
+	var dec decoder.DecoderType
```
vano144, line 276: Does the previous version look more readable?

vano144:
LGTM

pipeline/pipeline.go
anuriq: linter doesn't like it - ineffective assignment

ansakharov:
Good job!

PR #50, opened 2022-03-01
[feature-metrics] Add global package that handles the metrics

### LINKED ISSUE #49: Feature: add global package for metrics
Add a global module that would be used to submit metrics to Prometheus.

ssnd:
closes #49

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,4 +7,4 @@ WORKDIR /file.d
 
 COPY ./file.d .
 
-CMD ./file.d
+CMD [ "./file.d" ]
--- a/plugin/action/mask/mask.go
+++ b/plugin/action/mask/mask.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/ozontech/file.d/fd"
 	"github.com/ozontech/file.d/pipeline"
-	"github.com/prometheus/client_golang/prometheus"
+	"github.com/ozontech/file.d/stats"
 	insaneJSON "github.com/vitkovskii/insane-json"
 	"go.uber.org/zap"
 )
@@ -33,20 +33,17 @@ pipelines:
 }*/
 
 const (
-	substitution = byte('*')
-
-	metricName = "mask_plugin"
+	substitution    = byte('*')
+	metricSubsystem = "mask_plugin"
+	timesActivated  = "times_activated"
 )
 
-var MaskPromCounter = prometheus.NewCounter(prometheus.CounterOpts{})
-
 type Plugin struct {
-	config          *Config
-	sourceBuf       []byte
-	maskBuf         []byte
-	logMaskAppeared bool
-	valueNodes      []*insaneJSON.Node
-	logger          *zap.SugaredLogger
+	config     *Config
+	sourceBuf  []byte
+	maskBuf    []byte
+	valueNodes []*insaneJSON.Node
+	logger     *zap.SugaredLogger
 }
 
 //! config-params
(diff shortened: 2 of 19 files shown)


plugin/action/mask/mask.go
```diff
@@ -148,26 +145,18 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.ActionPluginP
 	p.valueNodes = make([]*insaneJSON.Node, 0)
 	p.logger = params.Logger
 	p.config.Masks = compileMasks(p.config.Masks, p.logger)
-	if p.config.MetricSubsystemName != nil {
```
ansakharov, line 151: Metric registration is required only if `MetricSubsystemName` is declared in config. This `if` can't be removed.

plugin/action/mask/mask.go
```diff
@@ -255,8 +244,9 @@ func (p *Plugin) Do(event *pipeline.Event) pipeline.ActionResult {
 		}
 		v.MutateToString(string(p.maskBuf))
 	}
-	if p.logMaskAppeared && maskApplied {
```
ansakharov, line 258: Same as above: `p.logMaskAppeared` describes does mask appearance needs to be sent by Prometheus.

stats/stats.go
```diff
@@ -0,0 +1,57 @@
...
+
+type stats struct {
+	allMetrics map[key]AnyMetric
+}
+
+var statsGlobal *stats
+
+func InitStats() {
+	statsGlobal = &stats{}
+	statsGlobal.allMetrics = make(map[key]AnyMetric)
```
ansakharov, line 30: Can do it shorter:
```go
statsGlobal = &stats{
    allMetrics: make(map[key]AnyMetric)
}
```

stats/stats.go
```diff
@@ -0,0 +1,58 @@
...
+		Name:      metricDesc.Name,
+		Help:      metricDesc.Help,
+	})
+
+	keyInternal := key{pipeline.PromNamespace, metricDesc.Subsystem, metricDesc.Name}
+	statsGlobal.registerMetric(keyInternal, maskPromCounter)
+}
+
+func GetCounter(subsystem, metricName string) prom.Counter {
+	return statsGlobal.allMetrics[getKey(subsystem, metricName)].(prom.Counter)
```
anuriq, line 47: panic won't happen in any case?

Dockerfile
```diff
@@ -7,4 +7,4 @@ WORKDIR /file.d
 
 COPY ./file.d .
 
-CMD ./file.d
+ENTRYPOINT [ "./file.d" ]
```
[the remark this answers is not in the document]
ssnd: I changed this to be able to pass command line arguments in docker-compose without changing the executable file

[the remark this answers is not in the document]
anuriq: i don't think this is needed. CMD can be overrided, and this is more expected than ENTRYPOINT

stats/stats.go
```diff
@@ -0,0 +1,86 @@
...
+		Help:      metricDesc.Help,
+	})
+
+	keyInternal := key{pipeline.PromNamespace, metricDesc.Subsystem, metricDesc.Name}
+	registerMetric(keyInternal, maskPromCounter)
+}
+
+func GetCounter(subsystem, metricName string) prom.Counter {
+	if statsGlobal == nil {
+		logger.Panicf("stats package uninitialized")
```
ansakharov, line 56: Is it responsibility of GetCounter? Probably better return error and resolve on higher level.

vano144: Another option is initialization in init function in stats package.

stats/stats.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	if val, ok := statsGlobal.allMetrics[getKey(subsystem, metricName)]; ok {
+		return val
+	}
+
+	return statsGlobal.allMetrics[getKey(subsystemName, unknownCounter)]
```
vano144, line 65: It can be calculated once

stats/stats.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type key struct {
+	namespace string
+	subsystem string
+	name      string
+}
+
+type stats struct {
+	allMetrics map[key]prom.Counter //
```
vano144, line 22: // ?

plugin/output/elasticsearch/elasticsearch.go
```diff
@@ -180,30 +205,39 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
...
 		root, err := insaneJSON.DecodeBytes(respContent)
 		if err != nil {
+			stats.GetCounter(subsystemName, sendErrorCounter).Inc()
 			p.logger.Errorf("wrong response from %s, will try other endpoint: %s", endpoint, err.Error())
 			insaneJSON.Release(root)
 			continue
 		}
 
 		if root.Dig("errors").AsBool() {
+			errors := 0.0
```
vano144, line 228: Why it is float ?

plugin/output/kafka/kafka.go
```diff
@@ -159,8 +175,9 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
 		for _, e := range errs {
 			p.logger.Errorf("can't write batch: %s", e.Err.Error())
 		}
+		stats.GetCounter(subsystemName, sendErrorCounter).Add(float64(len(errs)))
```
vano144, line 178: Why did you decide not to do a similar check here?
https://github.com/ozontech/file.d/pull/50/files#diff-ae9e68585530e0cefac2588a53be8e80e41ff31d9f37ae0cbf67bc1e971c45c5R237

stats/stats.go
ssnd: @anuriq @vano144 should we leave this panic here or remove it?

PR #56, opened 2022-03-23
Fuzzing

andrewmed:
Because of https://github.com/golang/go/issues/46312 no changes in Makefile yet, as I have not figured it out how to fuzz multiple packages

To run this one in kafka package, descend to directory and run 
`go test -tags=fuzz -fuzz .`

output should be smth like
Screenshot 2022-03-23 at 17 43 21

### DIFF
--- a/plugin/output/kafka/kafka_test.go
+++ b/plugin/output/kafka/kafka_test.go
@@ -1,0 +1,104 @@
+//go:build fuzz
+
+// To run: go test -tags=fuzz -fuzz .
+package kafka
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Shopify/sarama"
+	"github.com/ozontech/file.d/pipeline"
+	"github.com/stretchr/testify/require"
+	insaneJSON "github.com/vitkovskii/insane-json"
+	"go.uber.org/zap/zaptest"
+)
+
+const (
+	defaultTopic = "default_topic"
+	topicField   = "pipeline_kafka_topic"
+)
+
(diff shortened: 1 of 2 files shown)


plugin/output/kafka/kafka_test.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+type mockProducer struct {
+	t *testing.T
+}
+
+func (m *mockProducer) ensureTopic(msg *sarama.ProducerMessage) {
+	val, err := msg.Value.Encode()
+	if err != nil {
+		m.t.Fatal(err)
```
ansakharov, line 27: This verbose `err != nil` check can be replaced with `require.NoError(m.t, err)`.
require: https://pkg.go.dev/github.com/stretchr/testify/require#NoError

plugin/output/kafka/kafka_test.go
```diff
@@ -0,0 +1,105 @@
...
+		controller:   nil,
+		producer:     nil,
+		batcher:      nil,
+	}
+
+	f.Fuzz(func(t *testing.T, topicField, topicVal, key, val string) {
+		producer := mockProducer{
+			t: t,
+		}
+		p.producer = &producer
```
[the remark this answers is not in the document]
andrewmed: Done

plugin/output/kafka/kafka_test.go
andrewmed: Done

ssnd:
LGTM

PR #57, opened 2022-03-23
Postgres output plugin

ansakharov:
Pg output plugin.
TODO in this PR:

- [x] Tests
- [x]  Metrics 
- [x] Deduplications of different messages with same unique fields


plugin/output/postgres/query_builder copy.go
```diff
@@ -0,0 +1,131 @@
+package postgres
```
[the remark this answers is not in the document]
ansakharov: fixed

plugin/output/postgres/query_builder.go
```diff
@@ -0,0 +1,131 @@
...
+				postfix += ","
+			}
+		}
+		postfix += ") DO UPDATE SET "
+		for i, field := range updateableFields {
+			postfix += field + "=EXCLUDED." + field
+			if i < len(updateableFields)-1 {
+				postfix += ","
+			}
+		}
```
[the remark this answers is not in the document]
ansakharov: I will added comment as you wish, and move start of query to the const.
But part `postfix += field + "=EXCLUDED." + field` can't be templated.

Makefile
```diff
@@ -16,7 +16,7 @@ deps:
 
 .PHONY: cover
 cover:
-	go test -short -coverprofile=coverage.out ./...
+	go test -short -coverprofile=coverage.out ./plugin/output/postgres
```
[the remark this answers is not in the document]
andrewmed: why other cover is disabled?

plugin/output/postgres/query_builder.go
```diff
@@ -0,0 +1,140 @@
...
+		})
+		if col.Unique {
+			uniqueFields[col.Name] = colType
+		}
+	}
+
+	return pgFields, uniqueFields, nil
+}
+
+func (qb *pgQueryBuilder) createQuery(pgFields []column, table string) (sq.InsertBuilder, string) {
```
ssnd, line 104: imo it's confusing here that a function called `createQuery` returns both a query and a postfix

plugin/output/postgres/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+		p.logger.Fatalf("Invalid SQL. query: %s, args: %v, err: %v", query, args, err)
+		return
+	}
+
+	// Insert into pg with retry.
+	for i := retryCount; i > 0; i-- {
+		p.logger.Info(query, args)
+		_, err = p.db.Exec(query, args...)
+		if err != nil {
+			time.Sleep(time.Millisecond * 50)
```
ssnd, line 311: i wouldn't say it's necessary here, just an idea: maybe use a smarter retry strategy here? exponential backoff?

ssnd: anyway the time between retries shouldn't be 50ms i think

plugin/output/postgres/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+type PgxIface interface {
+	Exec(sql string, arguments ...interface{}) (commandTag pgx.CommandTag, err error)
+	Close() error
+}
+
+const (
+	outPluginType = "postgres"
+	subsystemName = "output_postgres"
+
+	retryCount = 3
```
ssnd, line 32: maybe move move this to the config?

plugin/output/postgres/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+				p.logger.Infof("event duplicated. Fields: %v, values: %v", pgFields, lVals)
+			} else {
+				uniqueEventsMap[uniqueID] = struct{}{}
+				atLeastOneGoodEvent = true
+				builder = builder.Values(lVals...)
+			}
+		}
+	}
+
+	builder = builder.Suffix(p.queryBuilder.GetPostfix()).PlaceholderFormat(sq.Dollar)
```
ssnd, line 294: why don't we move the query building logic inside the queryBuilder? looking for the first time you won't be able to understand what's going on anyway. we could create and return the sql from a hypothetic queryBuilder.GetSqlQuery and just handle the query building inside the module responsible for that

plugin/output/postgres/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+
+		// If LOOP_THROUGH_EVENT was breaked values mustn't be added.
+		if len(lVals) == len(pgFields) {
+			if _, ok := uniqueEventsMap[uniqueID]; ok {
+				stats.GetCounter(subsystemName, duplicatedEventCounter).Inc()
+				p.logger.Infof("event duplicated. Fields: %v, values: %v", pgFields, lVals)
+			} else {
+				uniqueEventsMap[uniqueID] = struct{}{}
+				atLeastOneGoodEvent = true
+				builder = builder.Values(lVals...)
```
ssnd, line 289: also we should start building the query only if we need to (e.g. atLeastOneGoodEvent==true).

plugin/output/postgres/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+			}
+		}
+	}
+
+	builder = builder.Suffix(p.queryBuilder.GetPostfix()).PlaceholderFormat(sq.Dollar)
+
+	// no valid events passed in non-strict mode.
+	if !atLeastOneGoodEvent {
```
ssnd, line 297: idea: maybe replace atLeastOneGoodEvent with haveNewEvents or smth similar

plugin/output/postgres/query_builder.go
ansakharov: Technically speaking it returns Builder and Postfix, while Query can be created by Builder.

plugin/output/postgres/postgres.go
ansakharov: 50ms is ok if network spiked or smth.
Exponential backoff is a good idea, we can move this values to config in future.
Now it looks good enough for me.

plugin/output/postgres/postgres.go
ansakharov: Good idea. Done.

plugin/output/postgres/postgres.go
ansakharov: `builder` mutated several times to this time and it differs from `p.queryBuilder.GetInsertBuilder()` now.

Is it much clearer in your opinion?
```go
func (qb *pgQueryBuilder) GetFinalBuilder(builder sq.InsertBuilder) sq.InsertBuilder  {
	return builder.Suffix(qb.postfix).PlaceholderFormat(sq.Dollar)
}

...
builder = p.queryBuilder.GetFinalBuilder(builder)
```

plugin/output/postgres/postgres.go
ansakharov: We can't declare `builder := p.queryBuilder.GetInsertBuilder()` inside `for _, event := range batch.Events` loop. Thats why be start create `builder` always

plugin/output/postgres/postgres.go
ansakharov: It's not about new events, but about event, that valid occur to config: events without required fields or with wrong type are invalid.
I'll rename it to `haveValidEvents`

plugin/output/postgres/postgres.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+func (p *Plugin) Out(event *pipeline.Event) {
+	p.batcher.Add(event)
+}
+
+func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
+	builder := p.queryBuilder.GetInsertBuilder()
+	pgFields := p.queryBuilder.GetPgFields()
+	uniqueFields := p.queryBuilder.GetUniqueFields()
```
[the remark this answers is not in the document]
ansakharov: Why not.

go.mod
```diff
@@ -96,3 +96,13 @@ require (
 	gopkg.in/yaml.v2 v2.2.8 // indirect
 	sigs.k8s.io/yaml v1.1.0 // indirect
 )
+
+require (
```
vano144, line 100: why is it a separate section ?

plugin/output/postgres/README.idoc.md
```diff
@@ -0,0 +1,5 @@
+# splunk HTTP Event Collector output
```
vano144, line 1: splunk ?

plugin/output/postgres/postgres.go
```diff
@@ -0,0 +1,385 @@
...
+	cancel()
+
+	if err != nil {
+		p.pool.Close()
+		p.logger.Fatalf("Failed insert into %s. query: %s, args: %v, err: %v", p.config.Table, query, args, err)
+	}
+}
+
+func (p *Plugin) processEvent(event *pipeline.Event, pgFields []column, uniqueFields map[string]pgType) (fieldValues []interface{}, uniqueID string, err error) {
+	fieldValues = make([]interface{}, 0, len(pgFields))
```
andrewmed, line 332: to avoid allocation here we can use same logic as in 
http://github.com/ozontech/file.d/blob/master/plugin/output/kafka/kafka.go#L138-L138 and pass slice to processEvent()


smth like
```
	if *workerData == nil {
		*workerData = &data{
			fieldValues = make([]interface{}, 0, len(pgFields))
		}
	}
```

here we will need just `fieldValues = fieldValues[:0]` instead

plugin/output/postgres/postgres.go
```diff
@@ -0,0 +1,357 @@
...
+				}
+				p.logger.Error(err)
+				stats.GetCounter(subsystemName, discardedEventCounter).Inc()
+				break LOOP_THROUGH_EVENT
+			}
+
+			lVals = append(lVals, lVal)
+
+			if _, ok := uniqueFields[field.Name]; ok {
+				if field.ColType == pgInt || field.ColType == pgTimestamp {
```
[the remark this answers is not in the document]
andrewmed: conversation moved here https://github.com/ozontech/file.d/pull/57#discussion_r840355920

vano144:
LGTM

PR #54, opened 2022-03-17
no need in unsafe in topic name

andrewmed:

PR #60, opened 2022-03-24
no need in unsafe in topic name - deep copying

andrewmed:
I prefer https://github.com/ozontech/file.d/pull/54/files over this one

### DIFF
--- a/pipeline/util.go
+++ b/pipeline/util.go
@@ -8,6 +8,16 @@ import (
 	"unsafe"
 )
 
+// Clone deeply copies string
+func CloneString(s string) string {
+	if len(s) == 0 {
+		return ""
+	}
+	b := make([]byte, len(s))
+	copy(b, s)
+	return *(*string)(unsafe.Pointer(&b))
+}
+
 // ByteToStringUnsafe converts byte slice to string without memory copy
 // This creates mutable string, thus unsafe method, should be used with caution (never modify provided byte slice)
 func ByteToStringUnsafe(b []byte) string {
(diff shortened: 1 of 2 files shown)


plugin/output/kafka/kafka.go
```diff
@@ -162,11 +162,7 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
 			data.messages[i] = &sarama.ProducerMessage{}
 		}
 		data.messages[i].Value = outBuf[start:]
-
-		// copy topic from json, to temporary out buffer to avoid event reusing issues
-		start = len(outBuf)
-		outBuf = append(outBuf, topic...)
-		data.messages[i].Topic = pipeline.ByteToStringUnsafe(outBuf[start:])
+		data.messages[i].Topic = pipeline.DeepCopyString(topic)
```
vano144, line 165: Let's just copy string only at 157 ? What do you think about it ?

pipeline/util.go
```diff
@@ -8,6 +8,14 @@ import (
 	"unsafe"
 )
 
+// DeepCopyString create new string with underlying buffer
+func DeepCopyString(s string) string {
```
vano144, line 12: https://go.googlesource.com/go/+/346fb7aaccda5cc059100502a02fe039ca0ea998/src/strings/clone.go ?

plugin/output/kafka/kafka.go
andrewmed: Good point! 
Done

plugin/output/kafka/kafka.go
```diff
@@ -154,19 +154,15 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
 		if p.config.UseTopicField {
 			fieldValue := event.Root.Dig(p.config.TopicField).AsString()
 			if fieldValue != "" {
-				topic = fieldValue
+				topic = pipeline.CloneString(topic)
```
vano144, line 157: pipeline.CloneString(fieldValue)

andrewmed: 🤦‍♂️
fixed

vano144:
LGTM

PR #68, opened 2022-04-11
CORELOG-179 splunk: better error handling

andrewmed:


plugin/output/splunk/splunk.go
```diff
@@ -208,8 +212,12 @@ func (p *Plugin) send(data []byte) error {
 		return fmt.Errorf("can't decode response: %w", err)
 	}
 
-	code := root.Dig("code").AsInt()
-	if code > 0 {
+	code := root.Dig("code")
```
vano144, line 215: AsInt() returns 0 if code if nil.

andrewmed: yes. that is why I made this PR)

andrewmed: there is difference between 0 error code and missing "code" field

vano144:
LGTM

Issue #67, opened 2022-04-11
Feature: add ban metric


ssnd:
Add a metric that shows if our service was banned by the [antispam plugin](https://github.com/ozontech/file.d/blob/master/pipeline/antispamer.go).

PR #73, opened 2022-04-13
Add gauge metric to the stats plugin

ssnd:
This PR adds gauge metric to the `stats` plugin. Gauge metric is a metric that represents some value that can arbitrarily go up and down. 
We also close the issue #67 by adding a metric that equals to 1 if file.d bans a specific sourceID because of spam.


pipeline/antispamer.go
```diff
@@ -78,11 +99,22 @@ func (p *antispamer) maintenance() {
 			logger.Infof("antispam: source has been unbanned id=%d", source)
 		}
 
+		if x >= p.threshold {
+			allUnbanned *= 0
```
andrewmed, line 103: why not just set sero?

stats/stats.go
```diff
@@ -20,13 +29,15 @@ type key struct {
...
 type stats struct {
-	allMetrics map[key]prom.Counter
+	counters map[key]prom.Counter
+	gauges   map[key]prom.Gauge
 }
 
 const (
 	subsystemName = "stats"
 
 	unknownCounter   = "unknown_counter"
```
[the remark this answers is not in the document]
ssnd: the stats package isn't critical for file.d, it works as an auxiliary module.  
that way, in case a metric wasn't initialised, the main system (file.d i/o & actions) shouldn't fail. that's why we added this "unknown" metric that basically gets incremented in case we try to increment an unregistered metric.

@vano144 @ansakharov what's your opinion on this?

andrewmed:
Looks like it is possible to measure the number of banned sources, not just one. Why would we not do it?

stats/stats.go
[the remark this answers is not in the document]
andrewmed: if a metric is not initialized at start and is requested, this is a bug 
current solution, on opposite, is a road to more subtle bugs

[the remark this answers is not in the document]
ansakharov: I see three ways:
- Return error: user (plugin in file.d) should decide how handle this err. But why should it decide? Not it's responsibility.
- Register metric on demand. But why it wasn't registered before? Something went wrong.
- Log this situation, probably increment other critical metric and fail fast.

@andrewmed suggested fail fast and it seems reasonable to me. Unregistered metric is fiasco. 
After all you can easily test if it registered: write simple test where you creating plugin and checking it's metric in `stats` pkg.

Neither you not I want spend time searching for phantom metric.

You probably would like take a look about `fail fast` conception.
For example: https://www.yegor256.com/2015/08/25/fail-fast.html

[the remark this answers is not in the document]
vano144: * You will not look at the logs when the application starts, that's why the metric is needed
* I don't think it's a reason to fall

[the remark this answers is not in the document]
ssnd: We decided to fix this issue later in a separate issue. 
cc @andrewmed @ansakharov

plugin/input/file/watcher_test.go
```diff
@@ -2,12 +2,13 @@ package file
 
 import (
 	"fmt"
-	"go.uber.org/atomic"
 	"os"
 	"path/filepath"
 	"testing"
 	"time"
 
+	"go.uber.org/atomic"
```
vano144, line 10: need join and sort imports here

ansakharov:
Add new version to Makefile please.

stats/stats.go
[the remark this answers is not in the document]
ansakharov: @vano144 it's reason to fail if we have proper tests, I'll suggest PR tomorrow.

stats/stats.go
```diff
@@ -39,49 +50,79 @@ func InitStats() {
...
 		return val
 	}
 
-	return statsGlobal.allMetrics[key{pipeline.PromNamespace, subsystem, metricName}]
+	return statsGlobal.counters[key{PromNamespace, subsystem, unknownCounter}]
 }
 
-func registerMetric(k key, metric prom.Counter) { // todo: support prom.Collector
+// Returns gauge for a given metric.
+func GetGauge(subsystem, metricName string) prom.Gauge {
```
vano144, line 98: Comments should be like // GetGauge returns gauge for a given metric.

stats/stats.go
```diff
@@ -39,49 +50,79 @@ func InitStats() {
...
+	if hasCounter || hasGauge {
 		logger.Errorf("rewriting existent metric")
 		GetCounter(subsystemName, duplicateCounter).Inc()
 	}
 
-	statsGlobal.allMetrics[k] = metric
+	switch mType {
+	case counterMetricType:
+		statsGlobal.counters[k] = metric.(prom.Counter)
+	case gaugeMetricType:
```
vano144, line 122: Do we need a case for unknown mType ?

vano144:
LGTM

plugin/input/file/watcher_test.go
ssnd: done

stats/stats.go
```diff
@@ -100,7 +146,7 @@ func (s *stats) registerOwnMetrics() {
 }
 
 func getKey(subsystem, metricName string) key {
-	return key{pipeline.PromNamespace, subsystem, metricName}
+	return key{PromNamespace, subsystem, metricName}
```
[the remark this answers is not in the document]
ssnd: fixed

stats/stats.go
ssnd: `registerMetric` is an internal function that will be called only from inside our module. we don't need to handle default cases because we have only two existent metric types that are added using the `RegisterCounter` and `RegisterGauge` functions.

pipeline/antispamer.go
```diff
@@ -52,6 +77,8 @@ func (p *antispamer) isSpam(id SourceID, name string, isNewSource bool) bool {
 	x := value.Inc()
 	if x == int32(p.threshold) {
 		value.Swap(int32(p.unbanIterations * p.threshold))
+		stats.GetGauge(subsystemName, antispamEnabled).Set(1)
```
ansakharov, line 80: Why `antispamEnabled` exists?

pipeline/pipeline.go
```diff
@@ -115,6 +115,7 @@ type Settings struct {
 
 // New creates new pipeline. Consider using `SetupHTTPHandlers` next.
 func New(name string, settings *Settings, registry *prometheus.Registry) *Pipeline {
+	fmt.Println("create new pipeline")
```
ansakharov, line 118: Forgotten debug statement. If not, change it to `logger.Info("creating new pipeline")`

stats/stats.go
ssnd: done

pipeline/antispamer.go
ssnd: changing the name to `antispamActive`

pipeline/pipeline.go
ssnd: sure, removing. thanks!

ansakharov (approved):
LGTM

PR #74, opened 2022-04-14
Add exponential backoff for pg output plugin

ansakharov:
# Description

Added expbackoff for pg output plugin while quering DB.


ssnd:
could you create and link an issue to this PR that describes the thing you're trying to fix. 
and also name the branches according to the guidelines we introduced a few days ago.

plugin/output/postgres/postgres.go
```diff
@@ -238,6 +240,17 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
+	stdBackoff := backoff.NewExponentialBackOff()
+	stdBackoff.Multiplier = 1.2
+	stdBackoff.RandomizationFactor = 0.25
+	stdBackoff.InitialInterval = p.config.Retention_
+	stdBackoff.MaxInterval = p.config.Retention_ * 2
+
+	ctxBackoff := backoff.WithContext(stdBackoff, p.ctx)
+	expBackoff := backoff.WithMaxRetries(ctxBackoff, uint64(p.config.Retry))
+
+	p.backoff = expBackoff
```
ssnd, line 252: maybe move this into another function?

plugin/output/splunk/splunk.go
```diff
@@ -98,6 +100,14 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
 	p.client = p.newClient(p.config.RequestTimeout_)
 
+	ctx := context.TODO()
+	stdBackoff := backoff.NewExponentialBackOff()
+	stdBackoff.Multiplier = 1.2
+	stdBackoff.RandomizationFactor = 0.25
+	stdBackoff.InitialInterval = time.Second
+	stdBackoff.MaxInterval = stdBackoff.InitialInterval * 2
+	p.backoff = backoff.WithContext(stdBackoff, ctx)
+
```
ssnd, line 110: same here, lots of magic constants. let's use a function for this

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/alecthomas/kingpin v2.2.6+incompatible
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d
 	github.com/bitly/go-simplejson v0.5.0
+	github.com/cenkalti/backoff/v4 v4.1.2
```
andrewmed, line 11: IMHO backoff does not need a dependency...

plugin/output/splunk/splunk.go
```diff
@@ -98,6 +100,14 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
 	p.config = config.(*Config)
 	p.client = p.newClient(p.config.RequestTimeout_)
 
+	ctx := context.TODO()
```
andrewmed, line 103: Background ?

plugin/output/postgres/postgres.go
```diff
@@ -65,6 +66,7 @@ type Plugin struct {
 	ctx        context.Context
 	cancelFunc context.CancelFunc
 
+	backoff      backoff.BackOff
```
andrewmed, line 69: do we need need this only to propagate calcellation to send() ?

how about taking it from elsewere?

for example from https://github.com/ozontech/file.d/blob/058ba19c99fcffdd6c6fef427adab5bc8e788bec/pipeline/batch.go#L62-L62

also, but it goes out of scope of this PR of course, setted way to use context is via params. better way to fix this would be imo passing it here http://github.com/ozontech/file.d/blob/master/pipeline/pipeline.go#L53-L53 but it is too much of a change right now

plugin/output/splunk/splunk.go
ansakharov: It should pass from pipeline in future, therefore TODO

plugin/output/postgres/postgres.go
ansakharov: I doesn't think it's a good idea, we can only move: 
```
stdBackoff := backoff.NewExponentialBackOff()
stdBackoff.Multiplier = 1.2
stdBackoff.RandomizationFactor = 0.25
```

Setting `InitialInterval`, `MaxInterval` and wrapping into `WithMaxRetries` can differ: Splunk doesn't have max retries and different logic of intervals. Seems like premature optimization for me.

plugin/output/splunk/splunk.go
ansakharov: Same as https://github.com/ozontech/file.d/pull/74#discussion_r859654027

plugin/output/postgres/postgres.go
ansakharov: We use it both for canellation and for backoff timer stretch during retry.

go.mod
ansakharov: We already have our bicycle: file.d. Let's avoid creating other one for exponential backoff.

plugin/output/postgres/postgres.go
andrewmed: Cancellation where? I cant find it

go.mod
andrewmed: Sure that is up to you, it is not blocking
I just think that  a new dependency for `t = t * 1.2` is too much

plugin/output/postgres/postgres.go
ansakharov: Ctx&retry cancellation sets here: https://github.com/ozontech/file.d/blob/b863c884c770d4c37713997b83d6c30f989ed715/plugin/output/postgres/postgres.go#L249-L250

andrewmed: yes that is what I am saying you create context and pass it in struct only to use later in backoff timer 
https://github.com/ozontech/file.d/blob/b863c884c770d4c37713997b83d6c30f989ed715/plugin/output/postgres/postgres.go#L350

you do not need it if you just check shouldStop()

cancellation in your commit is still done with another ctx
https://github.com/ozontech/file.d/blob/b863c884c770d4c37713997b83d6c30f989ed715/plugin/output/postgres/postgres.go#L339

andrewmed: Oh yes I can see now.
Well, I still think that two new contextes, passing it through struct and a new dependency is too much for a task that can be done just by a  for loop

ansakharov: We can do everything on programming with `for` loops. But it's not worth it.

plugin/output/postgres/postgres.go
```diff
@@ -238,6 +241,17 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
 	}
 	p.pool = pool
 
+	stdBackoff := backoff.NewExponentialBackOff()
```
vano144, line 244: maybe move this in function ? 
This will make it easier to change settings if needed.

vano144: also need fix conflicts in pr

ansakharov: done

VadimAlekseev (approved):
LGTM

plugin/output/postgres/postgres.go
```diff
@@ -241,6 +245,17 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
+	stdBackoff.Multiplier = consts.ExpBackoffDefaultMultiplier
+	stdBackoff.RandomizationFactor = consts.ExpBackoffDefaultRndFactor
+	stdBackoff.InitialInterval = p.config.Retention_
+	stdBackoff.MaxInterval = p.config.Retention_ * 2
+
+	ctxBackoff := backoff.WithContext(stdBackoff, p.ctx)
```
vitkovskii, line 254: If you use 3 variables they should be: 
`stdBackoff` > `expBackoff`
`ctxBackoff` > `expCtxBackoff`
`expBackoff` > `expCtxRetryBackoff`

But I'd recommend use just one  variable `backoff` and modify it.

expbackoff/backoff.go
```diff
@@ -0,0 +1,100 @@
...
+// New instance of BackoffTimeout.
+func New(
+	ctx context.Context,
+	counter prom.Counter,
+	timeout time.Duration,
+	retriesCfg RetriesCfg,
+	opts ...Option,
+) *BackOff {
+	var expBackoff backoff.BackOff
+	expBackoff = backoff.WithContext(new(opts...), ctx)
```
vitkovskii, line 33: expBackoff :=

expbackoff/backoff.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return &BackOff{
+		counter: counter,
+		backoff: expBackoff,
+		timeout: timeout,
+	}
+}
+
+// Exec processes gives lambda.
```
vitkovskii, line 45: given

plugin/output/postgres/postgres.go
```diff
@@ -306,33 +331,21 @@ func (p *Plugin) out(_ *pipeline.WorkerData, batch *pipeline.Batch) {
...
 		if err != nil {
-			outErr = err
-			p.logger.Infof("rows: %v, err: %s", rows, err.Error())
-			cancel()
-			time.Sleep(p.config.Retention_)
-			continue
-		} else {
-			outErr = nil
-			break
+			p.logger.Errorf("rows: %v, err: %s", rows, err.Error())
```
vitkovskii, line 342: use logging format similar to everywhere in the project

consts/consts.go
```diff
@@ -0,0 +1,6 @@
+package consts
+
+const (
+	ExpBackoffDefaultMultiplier = 1.2
```
vitkovskii, line 4: why we need new package here? use backoff package or move it to the pipeline settings

expbackoff/backoff.go
```diff
@@ -0,0 +1,100 @@
...
+func (b *BackOff) Exec(ctx context.Context, executor func(ctx context.Context) error) error {
```
vitkovskii, line 46: If BackOff is a wrapper, then use function names as it wrapped object

expbackoff/backoff.go
ansakharov: `var expBackoff backoff.BackOff` required here: `WithContext` and `WithMaxRetires` returns ptrs to different structs that implements same `Backoff` interface.

Otherwise it won't complile:
```
cannot use backoff.WithMaxRetries(expBackoff, retriesCfg.Limit) (value of type "github.com/cenkalti/backoff/v4".BackOff) as "github.com/cenkalti/backoff/v4".BackOffContext value in assignment: missing method Contextcompiler[InvalidIfaceAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal?utm_source%3Dgopls#InvalidIfaceAssign)
```

expbackoff/backoff.go
ansakharov: `RetryWithMetrics` seems appropriate.

backoff/backoff.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// New returns exponential backoff.
+func new(opts ...Option) *backoff.ExponentialBackOff {
+	backoff := backoff.NewExponentialBackOff()
+	for _, opt := range opts {
+		fmt.Println("op", opt)
+		opt(backoff)
+	}
+	fmt.Println(backoff)
```
vitkovskii, line 73: ????????

backoff/backoff.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil
+	}, b.backoff)
+	return err
+}
+
+// New returns exponential backoff.
+func new(opts ...Option) *backoff.ExponentialBackOff {
+	backoff := backoff.NewExponentialBackOff()
+	for _, opt := range opts {
+		fmt.Println("op", opt)
```
vitkovskii, line 70: ???????

plugin/output/postgres/postgres.go
```diff
@@ -123,14 +125,15 @@ type Config struct {
...
+	//> Timeouts can differ due using exponential backoff.
+	RequestTimeout  cfg.Duration `json:"db_request_timeout" default:"3000ms" parse:"duration"` //*
+	RequestTimeout_ time.Duration
 
 	//> @3@4@5@6
 	//>
 	//> Timeout for DB health check.
-	DBHealthCheckPeriod  cfg.Duration `json:"db_health_check_period" default:"60s" parse:"duration"` //*
-	DBHealthCheckPeriod_ time.Duration
+	HealthCheckPeriod  cfg.Duration `json:"db_health_check_period" default:"60s" parse:"duration"` //*
```
vitkovskii, line 135: interval

plugin/output/postgres/postgres.go
```diff
@@ -174,8 +177,14 @@ func (p *Plugin) registerPluginMetrics() {
 		Subsystem: subsystemName,
 		Help:      "Total pgsql duplicated messages",
 	})
+	stats.RegisterCounter(&stats.MetricDesc{
+		Name:      sendErrorCounter,
```
vitkovskii, line 181: use https://prometheus.io/docs/practices/naming/ naming guide

plugin/output/postgres/postgres.go
```diff
@@ -43,6 +43,7 @@ const (
 	// metrics
 	discardedEventCounter  = "event_discarded"
 	duplicatedEventCounter = "event_duplicated"
+	sendErrorCounter       = "send_error"
```
vitkovskii, line 46: request/out _total

also let's add label status="ok|error|retry|etc", instead of separate metrics for error

PR #78, opened 2022-04-19
New event metrics for pipelines

### LINKED ISSUE #77: Feature: add metrics for events count, size for input/output
**Is your feature request related to a problem? Please describe.**
`metrics_holder` only creates metrics for i/o events when there are actions between input&output. 
We want to be able to see the count and size of input (and output?) events even when there are no configured actions (i.e. there's only an input and output).

**Describe the solution you'd like**
Add a separate metric that shows input and output event count and size for each pipeline.

**Describe alternatives you've considered**
We could also create a dummy action plugin only to have the events count/size metric created though I don't think it's the right choice here.

ssnd:
# Description

Create new input/output event/size metrics for each pipeline.

Fixes #77


pipeline/pipeline.go
```diff
@@ -89,15 +96,17 @@ type Pipeline struct {
 
 	metricsHolder *metricsHolder
 
-	// some debugging shit
```
ansakharov, line 92: To offensive? :)
Looks funny to me

pipeline/pipeline.go
```diff
@@ -474,9 +514,34 @@ func (p *Pipeline) expandProcs() {
 	p.procCount.Swap(to)
 }
 
+// DeltaWrapper acts as a wrapper around int64
+// and returns the difference between
+// the new and the old value when a new value is set.
+// This is useful for metrics (see the maintenance function).
+type DeltaWrapper struct {
```
ansakharov, line 521: I suggest moving `DeltaWrapper` to other file better than declare at 521 line

pipeline/pipeline.go
```diff
@@ -474,9 +514,34 @@ func (p *Pipeline) expandProcs() {
...
+	val int64
+}
+
+func newDeltaWrapper() *DeltaWrapper {
+	return &DeltaWrapper{0}
+}
+
+func (c *DeltaWrapper) setAndGetDiff(newVal int64) int64 {
```
ansakharov, line 529: Weak naming, `DeltaWrapper` sets `val`, not `diff` which not obvious.
My suggestion:
```
// updateDelta sets new Delta and returns diff to previous Delta.
func (c *DeltaWrapper) updateDelta(val int64) (diff int64) {
	diff := newVal - c.val
	c.val = newVal
	return diff
}
```
Named return `diff` appropriate here, it describes returning int64.

And why receiver short name is `c`? `dw` seems better.

pipeline/pipeline.go
```diff
@@ -487,24 +552,31 @@ func (p *Pipeline) maintenance() {
...
+		deltaInputSize := inputSize.setAndGetDiff(p.inputSize.Load())
+		deltaOutputEvents := outputEvents.setAndGetDiff(p.outputEvents.Load())
+		deltaOutputSize := outputSize.setAndGetDiff(p.outputSize.Load())
 
-		rate := int(float64(deltaCommitted) * float64(time.Second) / float64(interval))
-		rateMb := float64(deltaSize) * float64(time.Second) / float64(interval) / 1024 / 1024
-
-		tc := totalCommitted
-		if totalCommitted == 0 {
-			tc = 1
```
ansakharov, line 501: This statement avoids division by zero.

pipeline/pipeline.go
```diff
@@ -487,24 +552,31 @@ func (p *Pipeline) maintenance() {
...
+		rate := int(float64(deltaInputEvents) * float64(time.Second) / float64(interval))
+		rateMb := float64(deltaInputSize) * float64(time.Second) / float64(interval) / 1024 / 1024
+		tc := int64(math.Max(float64(inputEvents.get()), 1))
+
+		p.logger.Infof(`%q pipeline stats interval=%ds, active procs=%d/%d, queue=%d/%d, out=%d|%.1fMb,`+
+			`rate=%d/s|%.1fMb/s, total=%d|%.1fMb, avg size=%d, max size=%d`,
+			p.Name, interval/time.Second, p.activeProcs.Load(), p.procCount.Load(),
+			p.settings.Capacity-p.eventPool.freeEventsCount, p.settings.Capacity,
+			deltaInputEvents, float64(deltaInputSize)/1024.0/1024.0, rate, rateMb,
+			inputEvents, float64(inputSize.get())/1024.0/1024.0, inputSize.get()/tc, p.maxSize)
```
ansakharov, line 579: Divizion by zero possibility: `inputSize.get()/tc`

pipeline/pipeline.go
```diff
@@ -474,9 +514,34 @@ func (p *Pipeline) expandProcs() {
...
+// DeltaWrapper acts as a wrapper around int64
+// and returns the difference between
+// the new and the old value when a new value is set.
+// This is useful for metrics (see the maintenance function).
+type DeltaWrapper struct {
+	val int64
+}
+
+func newDeltaWrapper() *DeltaWrapper {
+	return &DeltaWrapper{0}
```
ansakharov, line 526: 0 isn't required, just: `return &DeltaWrapper{}`

pipeline/pipeline.go
```diff
@@ -362,9 +403,8 @@ func (p *Pipeline) finalize(event *Event, notifyInput bool, backEvent bool) {
 
 	if notifyInput {
 		p.input.Commit(event)
-
-		p.totalCommitted.Inc()
-		p.totalSize.Add(int64(event.Size))
+		p.outputEvents.Inc()
+		p.outputSize.Add(int64(event.Size))
 
 		if len(p.outSample) == 0 && rand.Int()&1 == 1 {
```
ansakharov, line 409: I was sure that I commited it, but probably some PR wasn't merged.
Let's replace `if len(p.outSample) == 0 && rand.Int()&1 == 1` with `if len(p.outSample) == 0` to print always not in 50% of time. @vano144 are you ok with that?

pipeline/pipeline.go
```diff
@@ -89,15 +96,17 @@ type Pipeline struct {
...
 	eventLog        []string
 	eventLogMu      *sync.Mutex
 	inSample        []byte
 	outSample       []byte
-	totalCommitted  atomic.Int64
-	totalSize       atomic.Int64
+	outputEvents    atomic.Int64
+	outputSize      atomic.Int64
+	inputEvents     atomic.Int64
+	inputSize       atomic.Int64
```
ansakharov, line 109: Not big deal, but let's put `input` above `output`.

pipeline/pipeline.go
```diff
@@ -487,24 +552,31 @@ func (p *Pipeline) maintenance() {
...
+		deltaOutputSize := outputSize.setAndGetDiff(p.outputSize.Load())
 
-		rate := int(float64(deltaCommitted) * float64(time.Second) / float64(interval))
-		rateMb := float64(deltaSize) * float64(time.Second) / float64(interval) / 1024 / 1024
-
-		tc := totalCommitted
-		if totalCommitted == 0 {
-			tc = 1
+		// increment the metrics
+		getCounter := func(metric string) prometheus.Counter {
```
ansakharov, line 562: Is it required?
`stats.GetCounter(p.subsystemName(), metric).Add(float64(deltaOutputSize))` seems good.
And can be simplified to avoid `float64(deltaOutputSize)`.

pipeline/pipeline.go
```diff
@@ -487,24 +552,31 @@ func (p *Pipeline) maintenance() {
...
+			return stats.GetCounter(p.subsystemName(), metric)
 		}
-
-		p.logger.Infof("%q pipeline stats interval=%ds, active procs=%d/%d, queue=%d/%d, out=%d|%.1fMb, rate=%d/s|%.1fMb/s, total=%d|%.1fMb, avg size=%d, max size=%d", p.Name, interval/time.Second, p.activeProcs.Load(), p.procCount.Load(), p.settings.Capacity-p.eventPool.freeEventsCount, p.settings.Capacity, deltaCommitted, float64(deltaSize)/1024.0/1024.0, rate, rateMb, totalCommitted, float64(totalSize)/1024.0/1024.0, totalSize/tc, p.maxSize)
-
-		lastCommitted = totalCommitted
-		lastSize = totalSize
+		getCounter(inputEventsCountMetric).Add(float64(deltaInputEvents))
```
ansakharov, line 565: `float64(deltaInputEvents)` can be avoided if returning of `setAndGetDiff` float64

pipeline/pipeline.go
```diff
@@ -474,9 +514,34 @@ func (p *Pipeline) expandProcs() {
...
+}
+
+func newDeltaWrapper() *DeltaWrapper {
+	return &DeltaWrapper{0}
+}
+
+func (c *DeltaWrapper) setAndGetDiff(newVal int64) int64 {
+	diff := newVal - c.val
+	c.val = newVal
+	return diff
```
ansakharov, line 532: `return float64(diff)` to avoid afterward in: `getCounter(outputEventsCountMetric).Add(float64(deltaOutputEvents))`

pipeline/pipeline.go
andrewmed: Was thinking the same thing, I would prefer straight calls

pipeline/pipeline.go
ssnd: @vano144 said we should slowly get rid of this kind of stuff :-)

pipeline/pipeline.go
ssnd: done, thanks

pipeline/pipeline.go
ssnd: done

pipeline/pipeline.go
ssnd: done

pipeline/pipeline.go
ssnd: done

pipeline/pipeline.go
ssnd: done

pipeline/pipeline.go
vano144: It is only for debug  purpose. Why do you want change it? It is not about printing, it is about saving to outSample.

pipeline/pipeline.go
ssnd: I decided to name the function `updateValue` and use `diff` as the name for the returned value.  
That way we know we change the value (not the delta) and we get a `diff`erence between the new value and the old one.

pipeline/pipeline.go
ssnd: I'm not changing this then

pipeline/pipeline.go
ssnd: okay, done

pipeline/pipeline.go
ssnd: done, thank you

pipeline/pipeline.go
ansakharov: @vano144 it now printed only in 50% of times because `rand.Int()&1 == 1`. @vitkovskii proposed remove this.

pipeline/delta_wrapper.go
```diff
@@ -0,0 +1,27 @@
...
+	return &DeltaWrapper{}
+}
+
+// updateValue sets a new value for the wrapper
+// and **returns** the difference between
+// the current value and the new one.
+func (dw *DeltaWrapper) updateValue(newVal int64) (diff float64) {
+	diff = float64(newVal - dw.val)
+	dw.val = newVal
+	return
```
ansakharov, line 21: write `return diff` please. It's more stupid and easy.

pipeline/delta_wrapper.go
```diff
@@ -0,0 +1,27 @@
+package pipeline
+
+// DeltaWrapper acts as a wrapper around int64
+// and returns the difference between
+// the new and the old value when a new value is set.
+// This is useful for metrics (see the maintenance function).
+type DeltaWrapper struct {
```
ansakharov, line 7: Can you add simple test for `updateValue`, `get`?

pipeline/pipeline.go
```diff
@@ -3,7 +3,9 @@ package pipeline
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/ozontech/file.d/stats"
```
vano144, line 6: need sort imports

pipeline/pipeline.go
```diff
@@ -487,24 +530,27 @@ func (p *Pipeline) maintenance() {
...
+		stats.GetCounter(p.subsystemName(), outputEventsSizeMetric).Add(deltaOutputSize)
 
-		p.logger.Infof("%q pipeline stats interval=%ds, active procs=%d/%d, queue=%d/%d, out=%d|%.1fMb, rate=%d/s|%.1fMb/s, total=%d|%.1fMb, avg size=%d, max size=%d", p.Name, interval/time.Second, p.activeProcs.Load(), p.procCount.Load(), p.settings.Capacity-p.eventPool.freeEventsCount, p.settings.Capacity, deltaCommitted, float64(deltaSize)/1024.0/1024.0, rate, rateMb, totalCommitted, float64(totalSize)/1024.0/1024.0, totalSize/tc, p.maxSize)
+		rate := int(deltaInputEvents * float64(time.Second) / float64(interval))
+		rateMb := deltaInputSize * float64(time.Second) / float64(interval) / 1024 / 1024
+		tc := int64(math.Max(float64(inputEvents.get()), 1))
 
-		lastCommitted = totalCommitted
-		lastSize = totalSize
+		p.logger.Infof(`%q pipeline stats interval=%ds, active procs=%d/%d, queue=%d/%d, out=%d|%.1fMb,`+
```
vano144, line 548: maybe move this in function?

pipeline/pipeline.go
```diff
@@ -487,24 +530,27 @@ func (p *Pipeline) maintenance() {
 		p.antispamer.maintenance()
 		p.metricsHolder.maintenance()
 
-		totalCommitted := p.totalCommitted.Load()
-		deltaCommitted := int(totalCommitted - lastCommitted)
-
-		totalSize := p.totalSize.Load()
-		deltaSize := int(totalSize - lastSize)
+		// set new values and get the diffs
```
vano144, line 533: maybe move this in separate function?

pipeline/pipeline.go
```diff
@@ -487,24 +530,27 @@ func (p *Pipeline) maintenance() {
...
+		deltaOutputSize := outputSize.updateValue(p.outputSize.Load())
 
-		rate := int(float64(deltaCommitted) * float64(time.Second) / float64(interval))
-		rateMb := float64(deltaSize) * float64(time.Second) / float64(interval) / 1024 / 1024
-
-		tc := totalCommitted
-		if totalCommitted == 0 {
-			tc = 1
-		}
+		stats.GetCounter(p.subsystemName(), inputEventsCountMetric).Add(deltaInputEvents)
```
vano144, line 539: Can the delta be less than 0?

```
func (c *counter) Add(v float64) {
	if v < 0 {
		panic(errors.New("counter cannot decrease in value"))
	}
```

pipeline/pipeline.go
ssnd: done, thanks

pipeline/pipeline.go
ssnd: i added a check to the `update` function of the deltawrapper, we'll log an error if `diff < 0` and return abs(diff).

pipeline/pipeline.go
ssnd: done, please let me know if it looks okay

pipeline/pipeline.go
ssnd: done, please take a look before i resolve this

vano144:
LGTM

PR #79, opened 2022-04-29
Configurable logging, runtime level change

andrewmed:
Configurable logging, runtime level change

### DIFF
--- a/fd/file.d.go
+++ b/fd/file.d.go
@@ -265,6 +265,7 @@ func (f *FileD) startHTTP() {
 	mux.HandleFunc("/ready", f.serveLiveReady)
 	mux.HandleFunc("/freeosmem", f.serveFreeOsMem)
 	mux.Handle("/metrics", promhttp.Handler())
+	mux.Handle("/log/level", logger.Level)
 
 	f.server = &http.Server{Addr: f.httpAddr, Handler: mux}
 	longpanic.Go(f.listenHTTP)
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -2,30 +2,54 @@ package logger
 
 import (
 	"os"
+	"strings"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
-var Instance = zap.New(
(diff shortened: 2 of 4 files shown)


logger/logger.go
```diff
@@ -2,30 +2,54 @@ package logger
...
-		zapcore.AddSync(os.Stdout),
-		zap.NewAtomicLevelAt(zap.InfoLevel),
-	),
-).Sugar().Named("fd")
+var Instance *zap.SugaredLogger
+var Level zap.AtomicLevel
+
+func init() {
+	level := zap.InfoLevel
+	switch strings.ToLower(os.Getenv("LOG_LEVEL")) {
```
ssnd, line 16: I still think the initial debug level should be passed through command line options, not environment variables. Would love to hear @vano144 @ansakharov 's opinions on this.   
If we decide however to stick to your option, we should document it well. Now this just looks like some undocumented magic only file.d developers are aware of.

logger/logger.go
```diff
@@ -2,30 +2,54 @@ package logger
...
+	case "fatal":
+		level = zap.FatalLevel
+	}
+
+	Level = zap.NewAtomicLevelAt(level)
+
+	Instance = zap.New(
+		zapcore.NewCore(
+			zapcore.NewConsoleEncoder(zapcore.EncoderConfig{
+				// TimeKey:        "ts",
```
ssnd, line 34: ?

fd/file.d.go
```diff
@@ -265,6 +265,7 @@ func (f *FileD) startHTTP() {
 	mux.HandleFunc("/ready", f.serveLiveReady)
 	mux.HandleFunc("/freeosmem", f.serveFreeOsMem)
 	mux.Handle("/metrics", promhttp.Handler())
+	mux.Handle("/log/level", logger.Level)
```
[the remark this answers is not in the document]
andrewmed: Done

logger/logger.go
andrewmed: Done
Added to documentation

PR #81, opened 2022-04-29
Configurable logging, runtime level change 

andrewmed:
Configurable logging, runtime level change 

Fixes # 80

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -2,30 +2,54 @@ package logger
 
 import (
 	"os"
+	"strings"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
-var Instance = zap.New(
-	zapcore.NewCore(
-		zapcore.NewConsoleEncoder(zapcore.EncoderConfig{
-			// TimeKey:        "ts",
-			LevelKey:       "level",
-			NameKey:        "logger",
-			CallerKey:      "caller",
-			MessageKey:     "message",
-			StacktraceKey:  "stacktrace",
-			LineEnding:     zapcore.DefaultLineEnding,
-			EncodeLevel:    zapcore.LowercaseLevelEncoder,
-			EncodeTime:     zapcore.ISO8601TimeEncoder,
-			EncodeDuration: zapcore.SecondsDurationEncoder,
-			EncodeCaller:   zapcore.ShortCallerEncoder,
-		}),
-		zapcore.AddSync(os.Stdout),
-		zap.NewAtomicLevelAt(zap.InfoLevel),
-	),
-).Sugar().Named("fd")
+var Instance *zap.SugaredLogger
+var Level zap.AtomicLevel
+
+func init() {
+	level := zap.InfoLevel
+	switch strings.ToLower(os.Getenv("LOG_LEVEL")) {
+	case "debug":
+		level = zap.DebugLevel
+	case "info":
+		level = zap.InfoLevel
+	case "warn":
+		level = zap.WarnLevel
+	case "error":
+		level = zap.ErrorLevel
+	case "fatal":
+		level = zap.FatalLevel
+	}
+
+	Level = zap.NewAtomicLevelAt(level)
+
+	Instance = zap.New(
+		zapcore.NewCore(
+			zapcore.NewConsoleEncoder(zapcore.EncoderConfig{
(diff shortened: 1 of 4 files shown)


plugin/output/splunk/splunk.go
```diff
@@ -155,6 +156,7 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
 	data.outBuf = outBuf
 
 	for {
+		p.logger.Debugf("Trying to send: %s", string(outBuf))
```
vano144, line 159: Maybe it's better to write before the loop ?

andrewmed: Done!

Thanks for noticing

vano144:
LGTM

logger/logger.go
```diff
@@ -2,30 +2,58 @@ package logger
...
+	default:
+		level = defaultLevel
+	}
+
+	Level = zap.NewAtomicLevelAt(level)
+
+	Instance = zap.New(
+		zapcore.NewCore(
+			zapcore.NewConsoleEncoder(zapcore.EncoderConfig{
+				// TimeKey:        "ts",
```
ansakharov, line 38: Forgotten comment

plugin/output/splunk/splunk.go
```diff
@@ -146,13 +146,16 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
...
 		outBuf = root.Encode(outBuf)
 		_ = root.DecodeString("{}")
 	}
 	insaneJSON.Release(root)
 	data.outBuf = outBuf
+	
+	p.logger.Debugf("Trying to send: %s", outBuf)
 
 	for {
 		err := p.send(outBuf)
```
ansakharov, line 161: Probably it make sense to add `p.logger.Debugf("Failed to send: %s", outBuf)` if `if err != nil`

ansakharov:
Can you link your PR to issue?

logger/logger.go
andrewmed: it is in the source

plugin/output/splunk/splunk.go
andrewmed: It is already there

andrewmed: https://github.com/ozontech/file.d/pull/81/files#diff-b7d3945688c3a827787b7eeef8303aa7e0f7e5fe26fed4108068ad34f77e3301R164

andrewmed: Oh yes, the debug message, I will add it, thanks

andrewmed: Done

andrewmed:
@ansakharov issue is in the description

logger/logger.go
ansakharov: Let's delete from here and from source.

plugin/output/splunk/splunk.go
```diff
@@ -154,18 +155,22 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
...
 
+	p.logger.Debugf("Trying to send: %s", outBuf)
+
 	for {
 		err := p.send(outBuf)
 		if err != nil {
 			stats.GetCounter(subsystemName, sendErrorCounter).Inc()
-			p.logger.Errorf("can't send data to splunk address=%s: %s", p.config.Endpoint, err.Error())
+			p.logger.Errorf("Can't send data to splunk address=%s: %s", p.config.Endpoint, err.Error())
+			p.logger.Debugf("Error sending: %s", outBuf)
```
ansakharov, line 165: @vano144 described why this is a bad idea: infinite logs if splunk went down.  `Trying to send` is enough. So this `Debugf` is rendundant.

andrewmed: I thought this is what you said and it was a good idea

What infinite loop, could you explain more. This is logging of file.d instance

ansakharov: @vano144 described that if splunk went down `for` can iterates for very long time, with same debug message which is useless. That's why he suggests write debug msg once before `for` and once after.

ansakharov:
LGTM

PR #83, opened 2022-05-11
configurable masking

andrewmed:
Add configurable masking
Fixes https://github.com/ozontech/file.d/issues/84

### DIFF
--- a/plugin/action/mask/mask.go
+++ b/plugin/action/mask/mask.go
@@ -71,6 +71,16 @@ type Mask struct {
 	//>
 	//> Numbers of masking groups in expression, zero for mask all expression
 	Groups []int `json:"groups" required:"true"` //*
+
+	//> @3@4@5@6
+	//>
+	//> MaxCount limits the number of masked symbols in the masked output, if zero, no limit is set
+	MaxCount int `json:"max_count"` //*
+
+	//> @3@4@5@6
+	//>
+	//> ReplaceWord, if set, is used instead of asterisks for masking patterns that are of the same length or longer.
+	ReplaceWord string `json:"replace_word"` //*
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


plugin/action/mask/mask.go
```diff
@@ -85,11 +95,10 @@ func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
 }
 
 func compileMasks(masks []Mask, logger *zap.SugaredLogger) []Mask {
```
ansakharov, line 97: @vano144 in this plugins logger passes as arg while mostly used global logger. 
I guess we can use one approach.

plugin/action/mask/mask.go
```diff
@@ -162,19 +171,25 @@ func (p *Plugin) registerPluginMetrics() {
...
 }
 
-func appendMask(dst, src []byte, begin, end int) ([]byte, int) {
+func (p *Plugin) appendMask(mask *Mask, dst, src []byte, begin, end int) ([]byte, int) {
 	runeCounter := utf8.RuneCount(src[begin:end])
+	if mask.ReplaceWord != "" {
+		dst = append(dst, []byte(mask.ReplaceWord)...)
+		return dst, len(src[begin:end]) - len(mask.ReplaceWord)
+	}
 	for j := 0; j < runeCounter; j++ {
```
ansakharov, line 180: `j` usually used in nested cycles. `i` seems better here.

plugin/action/mask/mask.go
```diff
@@ -71,6 +71,16 @@ type Mask struct {
...
+
+	//> @3@4@5@6
+	//>
+	//> MaxCount limits the number of masked symbols in the masked output, if zero, no limit is set
+	MaxCount int `json:"max_count"` //*
+
+	//> @3@4@5@6
+	//>
+	//> ReplaceWord, if set, is used instead of asterisks for masking patterns that are of the same length or longer.
+	ReplaceWord string `json:"replace_word"` //*
```
ansakharov, line 83: Let's check in `Start` that `MaxCount` and `ReplaceWord` can't both appear.
Otherwise it's controversial config and we should throw fatal.

plugin/action/mask/mask.go
andrewmed: do not change this code

plugin/action/mask/mask.go
andrewmed: Done

Issue #92, opened 2022-05-23
Feature: Add 2 phase commit to s3 plugin


ansakharov:
** Issue **
2pc mechanism for s3 plugin.
Current implementation: s3 plugin writes to storage and commits.
Desired implementation: optional s3, that allows use 2pc in s3 and and other sources.

** Solution **
Create optional 2pc mechanism that commit to kafka.
Afterwards commits to other sources will be provided.
Kafka output plugin can be used to s3 plugin.

andrewmed:
Why S3 plugin needs to write something to another destination? AFAIK convention is that there is one output plugin for one destination

If this PR is not a 2-phase commit we should not be using this term

ansakharov:
I removed `2pc` term to avoid ambiguity.

PR #95, opened 2022-05-26
WIP

ansakharov:
Work in progress


andrewmed:
Why this is named a two phase commit? I can not find prepare - commit phases.
Why is a 2-phase commit is needed for pushing data to s3?

ansakharov:
> Why this is named a two phase commit? I can not find prepare - commit phases. Why is a 2-phase commit is needed for pushing data to s3?

Here is a bit of paraphrase here: 1 commit is while event writes to file and second when commit writes to kafka.

andrewmed:
Why S3 plugin needs to write something to another destination? AFAIK convention is that there is one output plugin for one destination

If this PR is not a 2-phase commit we should not be using this term

andrewmed:
There is also https://github.com/ozontech/file.d/issues/92 how it relates to this commit?

ansakharov:
> Why S3 plugin needs to write something to another destination? AFAIK convention is that there is one output plugin for one destination
> 
> If this PR is not a 2-phase commit we should not be using this term

Fair

ansakharov:
> Why S3 plugin needs to write something to another destination? AFAIK convention is that there is one output plugin for one destination
> 
> If this PR is not a 2-phase commit we should not be using this term

Writing to another destination required if one need records of s3 uploads and want escape integration with s3 storage. It's optional feature for those who need it.
About convention... it not that simple: s3 already uses multiple file plugins under it to save data in multiple buckets under one configuratoin.

plugin/output/s3/committer.go
```diff
@@ -0,0 +1,128 @@
...
+			ev.Root.AddFieldNoAlloc(ev.Root, i).MutateToNull()
+		case bool:
+			ev.Root.AddFieldNoAlloc(ev.Root, i).MutateToBool(values[i].(bool))
+		default:
+			logger.Infof("field %v skipped due unsupported type\n", i)
+		}
+	}
+
+	csw.commiter.Out(&ev)
+	insaneJSON.Release(ev.Root)
```
VadimAlekseev, line 85: You can replace this with defer statement

ansakharov: Good point

plugin/output/s3/s3.go
```diff
@@ -241,8 +147,19 @@ func Factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
 }
 
+// Start initializes and runs plugin. This function calls by pipeline and have hardcoded dependencies.
 func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginParams) {
-	p.StartWithMinio(config, params, p.minioClientsFactory)
+	committerFabric, err := NewCommitterFabric(config.(*Config), params)
+	if err != nil {
+		params.Logger.Fatal(err)
+	}
+	//var commFabric CommitterFabric
```
vadimalekseev, line 156: Do we need this commented code?

plugin/output/s3/s3.go
```diff
@@ -294,6 +212,18 @@ func (p *Plugin) StartWithMinio(config pipeline.AnyConfig, params *pipeline.Outp
 	p.uploadCh = make(chan fileDTO, p.config.FileConfig.WorkersCount_*4)
 	p.compressCh = make(chan fileDTO, p.config.FileConfig.WorkersCount_)
 
+	// Init and start commiter.
+	//if commFabric != nil && !commFabric.IsNil(commFabric) {
```
vadimalekseev, line 216: Do we need this commented code?

plugin/output/s3/s3.go
```diff
@@ -294,6 +212,18 @@ func (p *Plugin) StartWithMinio(config pipeline.AnyConfig, params *pipeline.Outp
...
+	//	if CommitterFabric(nil) != commFabric {
+	fmt.Println("is nil")
```
vadimalekseev, line 218: Debug print? :)

plugin/output/s3/s3.go
```diff
@@ -294,6 +212,18 @@ func (p *Plugin) StartWithMinio(config pipeline.AnyConfig, params *pipeline.Outp
...
+	if commFabric != nil { //&& !commFabric.IsNil(commFabric) {
+		fmt.Println("NOT NILL")
```
vadimalekseev, line 220: Same

plugin/output/s3/s3.go
```diff
@@ -495,14 +425,40 @@ func (p *Plugin) uploadWork() {
 		for {
 			err := p.uploadToS3(compressed)
 			if err == nil {
-				p.logger.Infof("successfully uploaded object: %s", compressed)
-				// delete archive after uploading
+				p.logger.Infof("successfully uploaded object: %v", compressed)
+				fmt.Println(p.config.CommitCfg)
+				// Send message to commiter if required.
```
vadimalekseev, line 430: comment in lowercase without a dot at the end of a sentence

plugin/output/s3/s3.go
ansakharov: yes 👍

vadimalekseev (approved):
LGTM

pipeline/event.go
```diff
@@ -42,10 +42,11 @@ const (
...
 
-	eventKindRegular int32 = 0
-	eventKindIgnore  int32 = 1
-	eventKindTimeout int32 = 2
-	eventKindUnlock  int32 = 3
+	eventKindRegular      int32 = 0
+	eventKindIgnore       int32 = 1
+	eventKindTimeout      int32 = 2
+	eventKindUnlock       int32 = 3
+	eventKindCustomCommit int32 = 4
```
andrewmed, line 49: iota will do nicely here

cfg/config.go
```diff
@@ -266,18 +266,24 @@ func Parse(ptr interface{}, values map[string]int) error {
 // 	}
 // }
 // this function will set `config.Child.T = config.T`
-// see file.d/cfg/config_test.go:TestHierarchy for an example
+// see file.d/cfg/config_test.go:TestHierarchy for an example.
 func ParseChild(parent reflect.Value, v reflect.Value, values map[string]int) error {
 	if v.CanAddr() {
-		for i := 0; i < v.NumField(); i++ {
-			name := v.Type().Field(i).Name
+		// Check for cfg pointers.
```
andrewmed, line 272: Not retated to your PR, but every time I look at this parsing it makes me sad. 
Why do we do black magic with something so simple as parsing a configuration  :(

plugin/output/s3/s3.go
```diff
@@ -495,14 +416,40 @@ func (p *Plugin) uploadWork() {
...
+					if p.config.CommitCfg.BucketNameFieldForTrack != "" {
+						commiterData[p.config.CommitCfg.BucketNameFieldForTrack] = compressed.bucketName
+					}
+					if p.config.CommitCfg.S3UrlNameFieldForTrack != "" {
+						commiterData[p.config.CommitCfg.S3UrlNameFieldForTrack] = compressed.fileName
+					}
+					err := p.commiterWrapper.CommitUpload(commiterData, p.config.CommitCfg.ConstantCommitMessagePart)
+					if err != nil {
+						p.logger.Fatalf("can't commit message of uploading: %s", err.Error())
+					}
```
andrewmed, line 439: Is this a desired logic - if we can not write S3 file meta once we just pass?
If so I think that `commit` is a wrong word. It is just writing meta smth like trySaveMeta()

cfg/config.go
```diff
@@ -266,18 +266,24 @@ func Parse(ptr interface{}, values map[string]int) error {
...
+		vIndirect := reflect.Indirect(v)
```
vano144, line 273: Need add new unit test, which shows a problem

plugin/output/s3/structs.go
```diff
@@ -0,0 +1,144 @@
...
+	BucketEventField string `json:"bucket_field_event" default:""` //*
+	//> @3@4@5@6
+	//> DynamicBucketsLimit regulates how many buckets can be created dynamically.
+	//> Prevents problems when some random strings in BucketEventField where
+	DynamicBucketsLimit int `json:"dynamic_buckets_limit" default:"32"` //*
+
+	// 2 phase commit section
+	//> @3@4@5@6
+	//> Describes config of commit.
+	CommitCfg *commitConfig `json:"commit_config" child:"true"` //*
```
vano144, line 143: Why is there a pointer? ?

plugin/output/file/timestamp/dates.go
```diff
@@ -0,0 +1,44 @@
+package timestamp
```
vano144, line 1: Here I suggest not to create a separate package for this. Just add this file to `plugin/output/file/`

plugin/output/file/timestamp/dates.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+// PrintFirstTimestamp returns first of timestamps.
+func (p Pair) PrintFirstTimestamp() uint64 {
+	return p.firstTimestamp.Load()
+}
+
+// PrintLastTimestamp returns last of timestamps.
+func (p Pair) PrintLastTimestamp() uint64 {
```
vano144, line 26: GetLastTimestamp

plugin/output/s3/committer.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/ozontech/file.d/plugin/output/kafka"
+	insaneJSON "github.com/vitkovskii/insane-json"
+)
+
+const (
+	// commiter type
+	KafkaType = "kafka"
+)
+
+var ErrKafkaInNonStrictMode = errors.New("can't use kafka as committer while pipeline settings mode isn't strict")
```
vano144, line 18: No usages

plugin/output/s3/committer.go
```diff
@@ -0,0 +1,116 @@
...
+		case nil:
+			ev.Root.AddFieldNoAlloc(ev.Root, i).MutateToNull()
+		case bool:
+			ev.Root.AddFieldNoAlloc(ev.Root, i).MutateToBool(values[i].(bool))
+		default:
+			logger.Infof("field %v skipped due unsupported type\n", i)
+		}
+	}
+
+	csw.commiter.Out(&ev)
```
vano144, line 85: But you return event to pool, is it work correct?

plugin/output/s3/s3.go
```diff
@@ -495,14 +416,40 @@ func (p *Plugin) uploadWork() {
...
+					if p.config.CommitCfg.LastTimestampFieldForTrack != "" {
+						commiterData[p.config.CommitCfg.LastTimestampFieldForTrack] = compressed.lastTimestamp
+					}
+					if p.config.CommitCfg.BucketNameFieldForTrack != "" {
+						commiterData[p.config.CommitCfg.BucketNameFieldForTrack] = compressed.bucketName
+					}
+					if p.config.CommitCfg.S3UrlNameFieldForTrack != "" {
+						commiterData[p.config.CommitCfg.S3UrlNameFieldForTrack] = compressed.fileName
+					}
+					err := p.commiterWrapper.CommitUpload(commiterData, p.config.CommitCfg.ConstantCommitMessagePart)
```
vano144, line 436: Can we create struct for commiterData ? and therefore CommitUpload may be simplified

plugin/output/s3/committer.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	switch config.CommitCfg.CommitterType {
+	case KafkaType:
+		return &KafkaCommitterFabric{config: config.CommitCfg.KafkaCfg, params: params}, nil
+	default:
+		return nil, fmt.Errorf("unknown commiter type: %s", config.CommitCfg.CommitterType)
+	}
+}
+
+// CommiterFunc returns func that starts plugin and returns it.
```
vano144, line 107: CommitterFunc

plugin/output/s3/committer.go
```diff
@@ -0,0 +1,116 @@
...
+func (cf *KafkaCommitterFabric) CommitterFunc() CommiterStarter {
+	anyPlugin, _ := kafka.Factory()
+	outPlugin := anyPlugin.(*kafka.Plugin)
```
vano144, line 110: how to stop it?

plugin/output/s3/structs.go
ansakharov: It's optional. Otherwise cfg parser will fail on default kafka plugin - on doesn't have broker list etc

cfg/config.go
ansakharov: It's optional in this case.

plugin/output/s3/committer.go
ansakharov: It's correct cuz we returns early in customCommit event type

cfg/config.go
ansakharov: Custom tags `child`, `slice` requires this

plugin/output/s3/s3.go
ansakharov: meta seems good

plugin/output/file/timestamp/dates.go
ansakharov: least and largest even better

plugin/output/s3/committer.go
ansakharov: Discussed and fixed

plugin/output/file/file.go
```diff
@@ -120,17 +178,37 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
...
 	}
 
 	p.idx = p.getStartIdx()
-	p.createNew()
+	p.startUpCreateNewFile()
 	p.setNextSealUpTime()
 
+	if p.config.MetaCfg.StoreMeta {
+		longpanic.Go(func() {
+			go func() {
```
andrewmed, line 192: why is go used here? longpanic.Go already runs func in a new goroutine

plugin/output/file/file.go
```diff
@@ -240,19 +375,27 @@ func (p *Plugin) sealUp() {
...
 	p.mu.Unlock()
 	if err := oldFile.Close(); err != nil {
 		p.logger.Panicf("could not close file: %s, error: %s", oldFile.Name(), err.Error())
 	}
-	logger.Errorf("sealing in %d, newFile: %s", time.Now().Unix(), newFileName)
+
+	logger.Errorf("sealing in %d, newFile: %s", time.Now().Unix(), finalOldFileName)
 	if p.SealUpCallback != nil {
-		longpanic.Go(func() { p.SealUpCallback(newFileName) })
+		longpanic.Go(func() { p.SealUpCallback(finalOldFileName, filepath.Base(outerPath)) })
```
andrewmed, line 398: would not just `longpanic.Go(p.SealUpCallback(..` work?

andrewmed (commented):
.

plugin/output/file/meta.go
```diff
@@ -0,0 +1,169 @@
+package file
```
andrewmed, line 1: how about adding tests for meta operations?

plugin/output/s3/s3_internals.go
```diff
@@ -107,10 +111,12 @@ func (p *Plugin) createOutPlugin(bucketName string) (*file.Plugin, error) {
 	anyPlugin, _ := file.Factory()
 	outPlugin := anyPlugin.(*file.Plugin)
 	outPlugin.SealUpCallback = p.addFileJobWithBucket(bucketName)
+	outPlugin.FileMetaCallback = p.genObjInfo(bucketName)
```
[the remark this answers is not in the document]
ansakharov: Good point, moved it under condition

plugin/output/file/meta.go
ansakharov: added into file_test.go

offset/json_offset.go
```diff
@@ -0,0 +1,48 @@
...
+func (o *jsonValue) Save(w io.Writer) error {
+	b, err := json.Marshal(o.value)
+	if err != nil {
+		return err
+	}
+	_, err = w.Write(b)
+	if err != nil {
+		return err
+	}
+	_, err = w.Write([]byte("\n"))
```
vano144, line 29: Please add a comment about it.

offset/json_offset.go
```diff
@@ -0,0 +1,48 @@
...
+	return nil
+}
+
+func newJsonOffset(path string, value interface{}) *Offset {
+	res := NewOffset(path)
+	res.Callback = &jsonValue{value}
+	return res
+}
+
+func LoadJson(path string, value interface{}) error {
```
vano144, line 42: Why is it unused func ?

plugin/output/file/file.go
```diff
@@ -107,6 +133,38 @@ func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.OutputPluginP
 	p.fileName = file[0 : len(file)-len(p.fileExtension)]
 	p.tsFileName = "%s" + "-" + p.fileName
 
+	ctx, cancel := context.WithCancel(context.Background())
+	p.cancel = cancel
+
+	if p.config.MetaCfg.StoreMeta {
+		metaWriter, err := newMeta(
```
vano144, line 140: Seems like it would be better to add a function constructor for meta

plugin/output/file/file.go
```diff
@@ -209,24 +310,53 @@ func (p *Plugin) write(data []byte) {
...
 
-func (p *Plugin) createNew() {
-	p.tsFileName = fmt.Sprintf("%d%s%s%s", time.Now().Unix(), fileNameSeparator, p.fileName, p.fileExtension)
+// startUpCreateNewFile creates of restores log and meta files during startUp
+func (p *Plugin) startUpCreateNewFile() {
+	p.createNew("", "")
+}
+
+// createNew creates new or appends to existing log file
+func (p *Plugin) createNew(sealingLogFile, sealingOuterPath string) {
```
vano144, line 319: Let's return an error here and call fatal in one place.

plugin/output/file/file.go
```diff
@@ -240,19 +370,30 @@ func (p *Plugin) sealUp() {
...
 	oldFile := p.file
 	p.mu.Lock()
-	p.createNew()
+
+	outerPath, ext := "", ""
+	if p.FileMetaCallback != nil {
+		outerPath, ext = p.FileMetaCallback(filepath.Base(finalOldFileName))
+	}
+
+	p.createNew(filepath.Base(finalOldFileName)+ext, outerPath)
```
vano144, line 384: Better to use fmt Sprintf. This will be more readable.

plugin/output/file/file.go
```diff
@@ -169,12 +244,38 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
 
 	outBuf := data.outBuf[:0]
 
+	minTs, maxTs := int64(0), int64(0)
 	for _, event := range batch.Events {
 		outBuf, _ = event.Encode(outBuf)
 		outBuf = append(outBuf, byte('\n'))
+
+		if p.config.MetaCfg.StoreMeta {
+			ts, err := event.Root.DigStrict("timestamp")
```
vano144, line 253: TimestampField ?

plugin/output/file/file.go
```diff
@@ -169,12 +244,38 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
...
+			if err != nil {
+				p.logger.Errorf("doesn't have timestamp field: %s", err.Error())
+				continue
+			}
+			tsInt, err := ts.AsInt64()
```
vano144, line 258: How do we support other formats ?

plugin/output/file/file.go
```diff
@@ -169,12 +244,38 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
...
+			}
+			if tsInt > maxTs {
+				maxTs = tsInt
+			}
+		}
 	}
 	data.outBuf = outBuf
 
+	if p.config.MetaCfg.StoreMeta {
+		p.mu.Lock()
```
vano144, line 274: Please add comment about need for blocking

plugin/output/file/file_test.go
```diff
@@ -138,6 +141,127 @@ func TestSealUpHasContent(t *testing.T) {
...
+			sealedPrefix:            p.config.MetaCfg.SealedMetaPrefix,
+			sealedFileFieldName:     p.config.MetaCfg.SealedFileNameField,
+			sealedFilePathFieldName: p.config.MetaCfg.SealedFilePathFieldName,
+			fileMode:                p.config.FileMode_,
+			staticMeta:              p.config.MetaCfg.StaticMeta,
+		},
+	)
+	require.NoError(t, err)
+	now := time.Now().Unix()
+	metaWriter.newMetaFile(metaWriter.fileName, sealUpDTO{}, now)
```
vano144, line 211: Unhandled error

plugin/output/file/meta.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+const (
+	metaFilePrefix      = "meta_"
+	firstTimestampField = "first_timestamp"
+	lastTimestampField  = "last_timestamp"
+)
+
+type meta struct {
+	filePreifx,
```
vano144, line 21: filePrefix, extension

plugin/output/file/file.go
ansakharov: Doesn't support now.

plugin/output/file/file_test.go
ansakharov: Now both places checked.

plugin/output/file/file.go
ansakharov: Good poiunt.

plugin/output/file/file.go
```diff
@@ -83,6 +86,26 @@ type Config struct {
...
+
+	//> MetaCfg describes metafile with sealfile desription
+	MetaCfg MetaConfig `json:"meta_config" child:"true"`
+}
+
+type MetaConfig struct {
+	//> Stores metadata of sealed files.
+	StoreMeta bool `json:"store_meta"` //*
+	//> MetaDataFile saves metainformation about sealed file.
+	MetaDataDir string `json:"metadata_dir" default:"/var/log/file-d.meta"` //*
```
vitkovskii, line 98: Wrong comment, should be for `MetaDataDir`.  Also why don't do like in the original file config, just define full filename  (with path, etc) and some layout. https://github.com/ozontech/file.d/blob/master/plugin/output/file/file.go#L62 https://github.com/ozontech/file.d/blob/master/plugin/output/file/file.go#L69

plugin/output/file/file.go
```diff
@@ -83,6 +86,26 @@ type Config struct {
...
+	//> SealedMetaPrefix defines prefix of sealed meta file
+	SealedMetaPrefix string `json:"sealed_metafile_prefix" default:"sealed_meta_"` //*
+	//> TimestampField that contains valid timestamp.
+	TimestampField string `json:"timestamp_field" default:"timestamp"` //*
```
vitkovskii, line 102: is it field name of json where timestamp will be stored?

plugin/output/file/file.go
```diff
@@ -83,6 +86,26 @@ type Config struct {
...
+	//> Static part of json meta.
+	StaticMeta string `json:"static_meta" default:""` //*
```
vitkovskii, line 104: what is static part of json?

plugin/output/file/file.go
```diff
@@ -83,6 +86,26 @@ type Config struct {
...
+	//> SealedFileNameField stores name of sead file
+	SealedFileNameField string `json:"sealed_filename_field" default:"sealed_name"` //*
```
vitkovskii, line 106: why we need separate file name and path?

vitkovskii (changes requested):

plugin/output/file/file.go
```diff
@@ -169,12 +235,43 @@ func (p *Plugin) out(workerData *pipeline.WorkerData, batch *pipeline.Batch) {
...
+
+		if p.config.MetaCfg.EnableMetaFiles {
+			ts, err := event.Root.DigStrict(p.config.MetaCfg.TimestampField)
+			if err != nil {
+				p.logger.Errorf("doesn't have timestamp field: %s", err.Error())
+				continue
+			}
+			tsInt, err := ts.AsInt64()
+			if err != nil {
+				p.logger.Errorf("fime field isn't int64: %s", err.Error())
```
vitkovskii, line 251: typo

PR #104, opened 2022-06-02
Fix codestyle issues

### LINKED ISSUE #103: Feature: improve code style
**Is your feature request related to a problem? Please describe.**
File.d codebase contains other code styles.

**Describe the solution you'd like**
Configure the golangci-lint and fix common styling issues.

vadimalekseev:
# Description

This configures the linter and fixes errors that are not checked by the linter.

Fixes #103


ansakharov (approved):
LGTM.
Good description of linters.

plugin/input/http/http.go
```diff
@@ -246,7 +245,7 @@ func (p *Plugin) serve(w http.ResponseWriter, r *http.Request) {
 
 	_, err := w.Write(result)
 	if err != nil {
-		stats.GetCounter(subsystemName, httpErrorCounter).Inc()
+		metrics.GetCounter(subsystemName, httpErrorCounter).Inc()
```
andrewmed, line 248: is renaming of package related to linter?
I just try to see net effect of using linter in your configuration, and what I see is just this package renaming...

vadimalekseev:
@andrewmed 

> Did you try it running? It gave me following:

gci linter is not included. Take the actual changes and try again

> why `pipeline/pipeline.go:424:31: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec) if len(p.outSample) == 0 && rand.Int()&1 == 1 {` ???????? if we want crypto we use it if not, we dont

If a crypto-resistant random is not needed here, you can write this:
```
//nolint:gosec
```
Or we can disable this check globally.

> or `plugin/output/postgres/postgres.go:261:4: ifElseChain: rewrite if-else to switch statement (gocritic) if errors.Is(err, ErrEventDoesntHaveField) {` why we would be failing tests on this subjective view ???

What's the problem with switch-case? We can disable this check.

> `` cfg/config.go:231:18: string `true` has 3 occurrences, make it a constant (goconst) if childTag == "true" { `` does this check takes into account that these three occurences might be unrelated?

These 3 cases are related, it's a hardcoded string, I don't see any problem with making it into a constant.

> `` plugin/output/splunk/splunk.go:176:30: ptrToRefParam: consider `workerData' to be of non-pointer type (gocritic) func (p *Plugin) maintenance(workerData *pipeline.WorkerData) {} `` why on earth linter fails tests on seeing a pointer?

This linter complains that you are using a reference to a reference structure. For example **int, *interface{}, *[]int, etc. We can disable this check for the entire package, if so required by our logic.


I collected and looked at all the changes that the linter wants to make. I don't see unnecessary rules.
The linters in this PR are a very common set used by many projects. For example, [filebeat](https://github.com/elastic/beats/blob/a451e35c07329f2810e85c047b42a70a74dd4893/.golangci.yml#L35-L61)

vitkovskii (approved):

ansakharov (approved):
LGTM

vano144 (approved):

PR #119, opened 2022-07-06
Fix max_event_size

### LINKED ISSUE #118: Bug: max_log_size doesn't work
**Describe the bug**
Events stop processing if at least one event exceeds the size of max_event_size

Version: 0.5.10

vadimalekseev:
Fixes #118


plugin/input/file/worker_test.go
```diff
@@ -108,7 +115,145 @@ func TestWorkerWork(t *testing.T) {
...
+				`{"a":"a"}` + "\n",
+			},
+		},
+		{
+			name:           "no new line",
+			maxEventSize:   50,
+			readBufferSize: 1024,
+
+			inData:  `{"a":"a"}`,
+			outData: nil, // TODO: it is ok? we don't send an event if we don't find a newline
```
vadimalekseev, line 206: @vano144

vadimalekseev: we can fix this easy:
```
pos := int64(bytes.IndexByte(buf, '\n'))

// if it is the end of the file
+ if len(buf) > n && pos == -1 {
+ 	pos = n - 1
+}

if pos == -1 {
	scanned += int64(len(buf))
	break
}
```

vano144: it is ok

PR #125, opened 2022-07-07
122-added-extra-field-to-mask-plugin

### LINKED ISSUE #122: Feature: mask action add field
**Issue**
Sometime it's hard to find masked event in storage. Need solution, that helps filter such events.

**Solution**
Add into config optional props: masked_key, masked_value.
If masked_key were set, than ```masked_key: masked_value``` should be added to event.

ansakharov:
Extra key-val added to event if mask appers.

Fixes #122  (issue)

### DIFF
--- a/plugin/action/mask/mask.go
+++ b/plugin/action/mask/mask.go
@@ -51,7 +51,7 @@ type Plugin struct {
 type Config struct {
 	//> @3@4@5@6
 	//>
-	//> If set counterMetric with this name would be sent on metric_subsystem_name.mask_plugin
+	//> If set counterMetric with this name would be sent on metric_subsystem_name.mask_plugin.
 	MetricSubsystemName *string `json:"metric_subsystem_name" required:"false"` //*
 
 	//> @3@4@5@6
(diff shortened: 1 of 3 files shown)


plugin/action/mask/mask.go
```diff
@@ -273,6 +282,9 @@ func (p *Plugin) Do(event *pipeline.Event) pipeline.ActionResult {
 		stats.GetCounter(*p.config.MetricSubsystemName, timesActivated).Inc()
 		p.logger.Infof("mask appeared to event, output string: %s", event.Root.EncodeToString())
 	}
+	if p.config.MaskedEventExtraField != "" && maskApplied {
```
vano144, line 285: Let's move this before logging (281) to see result in logs

vano144 (approved):

plugin/action/mask/mask.go
ansakharov: It will help to avoid inconvenience.

plugin/action/mask/mask.go
```diff
@@ -58,6 +58,15 @@ type Config struct {
 	//>
 	//> List of masks.
 	Masks []Mask `json:"masks"` //*
+
+	//> @3@4@5@6
+	//>
+	//> if set than MaskedEventExtraField: MaskedEventExtraValue will be written to event
```
vadimalekseev: Upper case for 'if' :)

vadimalekseev (approved):

PR #152, opened 2022-08-08
Add join_template action plugin with go_panic template

### LINKED ISSUE #151: Feature: Add join go panic plugin with predefined regex
Add join go panic plugin as an alias to "join" plugin with predefined `start` and `continue`regex.

HeadHunter483:
Closes #151


vadimalekseev (approved):

plugin/action/join_go_panic/join_go_panic_test.go
```diff
@@ -0,0 +1,263 @@
...
+				inEvents.Inc()
+			})
+
+			outEvents := atomic.Int32{}
+			lastID := atomic.Uint64{}
+			output.SetOutFn(func(e *pipeline.Event) {
+				outEvents.Inc()
+				id := lastID.Swap(e.SeqID)
+				if id != 0 && id >= e.SeqID {
+					panic("wrong id")
```
ansakharov, line 234: Why panic? `require.False(t, id != 0 && id >= e.SeqID)` seems better

plugin/action/join_go_panic/join_go_panic_test.go
```diff
@@ -0,0 +1,263 @@
...
+				}
+				lines = append(lines, fmt.Sprintf(format, line))
+			}
+
+			config := &Config{
+				Field: "log",
+			}
+
+			err := cfg.Parse(config, nil)
+			if err != nil {
```
ansakharov, line 208: require.NoError(t, err)

plugin/action/join_go_panic/join_go_panic.go
```diff
@@ -0,0 +1,96 @@
+package join_go_panic
```
ansakharov, line 1: You decided do separate plugins for go, java etc? It can be one plugin with different modes. But one plugin won't be simple

plugin/action/join_go_panic/join_go_panic_test.go
vadimalekseev: It's copy-paste :)

HeadHunter483: Исправил.

plugin/action/join_go_panic/join_go_panic.go
HeadHunter483: Yes, I think  it's a good idea to create one plugin for all such cases, where another languages patterns could be added as new options. So I propose an implementation of [join_template plugin](https://github.com/ozontech/file.d/pull/152/files#diff-d49806fa9cd1fad1ddc6ed32bb2e74747f59eaf9428a64c01eb65f35517a21cc).

vano144 (approved):

PR #156, opened 2022-08-09
redis throttle done

### LINKED ISSUE #82: Feature: redis throttle
For better separation of throttling required different mechanism of limiting.

**Solution**
Use Redis to store limit for each pod and renew it. In absence of Redis fallback to memory limiter.

ansakharov:
Fixes #82


plugin/action/throttle/redis_limiter.go
```diff
@@ -0,0 +1,179 @@
...
+	maxID := l.incrementLimiter.rebuildBuckets(n)
+	_ = l.totalLimiter.rebuildBuckets(n)
+
+	minID := l.totalLimiter.minID
+	count := l.incrementLimiter.bucketCount
+
+	// get actual bucket value
+	keyPrefixLen := l.keyPrefix.Len()
+
+	keyIdxs := make([]int, 0, count)
```
vitkovskii, line 86: let's allocate only once on constructor

vitkovskii (changes requested):

plugin/action/throttle/throttle.go
```diff
@@ -104,38 +170,119 @@ func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
+			throttleField,
+			limiterKey,
+			p.config.BucketInterval_,
+			p.config.BucketsCount,
+			rule.limit,
+		)
+	case inMemoryBackend:
+		return NewInMemoryLimiter(p.config.BucketInterval_, p.config.BucketsCount, rule.limit)
+	default:
+		p.logger.Panicf("unknown limiter kind: %s", p.config.LimiterBackend)
```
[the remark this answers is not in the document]
ansakharov: thx fixed.

plugin/action/throttle/redis_limiter.go
ansakharov: thx fixed.

plugin/action/throttle/throttle.go
```diff
@@ -102,40 +168,135 @@ func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
+
+		ticker := time.NewTicker(p.config.RedisBackendCfg.SyncInterval_)
+
+		// run syncer at single throttle
+		go limiterSync.Do(func() {
+			jobs := make(chan limiter, p.config.RedisBackendCfg.WorkerCount)
+			jobsDone := make(chan struct{}, p.config.RedisBackendCfg.WorkerCount)
 
-	p.rules = append(p.rules, NewRule(map[string]string{}, complexLimit{p.config.DefaultLimit, p.config.LimitKind}))
+			for i := 0; i < p.config.RedisBackendCfg.WorkerCount; i++ {
```
ansakharov, line 227: This workers append only once for pipeline.

plugin/action/throttle/throttle.go
```diff
@@ -102,40 +168,135 @@ func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
+			for {
+				select {
+				case <-ctx.Done():
+					close(jobs)
+					return
+				case <-ticker.C:
+					limitersMu.RLock()
+					jobsCount := 0
+
+					for _, lim := range limiters[p.pipeline] {
```
ansakharov, line 240: Get limit only for current pipeline

plugin/action/throttle/redis_limiter.go
```diff
@@ -0,0 +1,181 @@
...
+	rl.keyPrefix.WriteString("_")
+
+	return rl
+}
+
+func (l *redisLimiter) isAllowed(event *pipeline.Event, ts time.Time) bool {
+	// count increment from last sync
+	isAllowed := l.incrementLimiter.isAllowed(event, ts)
+	if isAllowed {
+		// count global limiter if distributed mode on and passed local limit
```
vitkovskii, line 64: where is no "distributed" mode in the settings, make comment actual

plugin/action/throttle/throttle.go
```diff
@@ -102,40 +168,145 @@ func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
+						return
+					case <-ticker.C:
+						limitersMu.RLock()
+						jobsCount := 0
+
+						for _, lim := range limiters[p.pipeline] {
+							jobs <- lim
+							jobsCount++
+						}
+						for i := 0; i < jobsCount; i++ {
```
vitkovskii, line 249: use wait group here

plugin/action/throttle/throttle.go
```diff
@@ -1,32 +1,59 @@
...
 )
 
 var (
 	defaultThrottleKey = "default"
 
 	// limiters should be shared across pipeline, so let's have a map by namespace and limiter name
-	limiters   = map[string]map[string]*limiter{} // todo: cleanup this map?
+	limiters   = map[string]map[string]limiter{} // todo: cleanup this map?
 	limitersMu = &sync.RWMutex{}
+	// only one throttle plugin updates limiters.
```
vitkovskii, line 21: what does it mean?))

vitkovskii (changes requested):

plugin/action/throttle/redis_limiter.go
ansakharov: now it's really fixed.

plugin/action/throttle/throttle.go
ansakharov: It's artifacts. I'm deleting it.

ansakharov: Deleted.

plugin/action/throttle/throttle.go
ansakharov: Added

plugin/action/throttle/redis_limiter.go
ansakharov: Good point

plugin/action/throttle/redis_limiter.go
ansakharov: now it's fixed.

plugin/action/throttle/redis_limiter.go
```diff
@@ -0,0 +1,188 @@
...
+		l.bucketIdsForSync = append(l.bucketIdsForSync, minID+i)
+	}
+
+	l.totalLimiter.mu.Unlock()
+	l.incrementLimiter.mu.Unlock()
+
+	l.syncLocalGlobalLimiters(maxID)
+	l.updateKeyLimit(keyPrefixLen)
+
+	l.keyIdxsForSync = l.keyIdxsForSync[:0]
```
vitkovskii, line 111: it's more safety to clear slice before using it

plugin/action/throttle/throttle.go
```diff
@@ -85,6 +121,35 @@ type Config struct {
 	Rules []RuleConfig `json:"rules" default:"" slice:"true"` //*
 }
 
+type RedisBackendConfig struct {
+	//> @3@4@5@6
+	//>
+	// Host:Port of redis server.
+	Host string `json:"host"` //*
```
vitkovskii, line 128: It's better to call it `endpoint` like in other plugins. Description should be: "Аddress of redis server. Format: `HOST:PORT`

plugin/action/throttle/throttle.go
```diff
@@ -102,40 +167,143 @@ func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
+		if pingResp := p.redisClient.Ping(); pingResp.Err() != nil {
+			p.logger.Fatalf("can't ping redis: %s", pingResp.Err())
+		}
+
+		ticker := time.NewTicker(p.config.RedisBackendCfg.SyncInterval_)
+
+		limitersMu.Lock()
+		wg := sync.WaitGroup{}
+		// run syncer at single throttle for one pipeline
+		if _, ok := limiterSyncMap[p.pipeline]; !ok {
```
vitkovskii, line 224: It will not work, many goroutines may pass into this condition before `limiterSyncMap[p.pipeline] = struct{}{}`. Use `sync.Once()`

vitkovskii (changes requested):

plugin/action/throttle/throttle.go
ansakharov: They won't pass `limitersMu.Lock()` few lines above

ansakharov: Actual flow: 
```
limiterMu.Lock()
if _, ok := limiterSyncMap[p.pipeline]; !ok {
     limiterSyncMap[p.pipeline] = struct{}{}
     limiterMu.Unlock()
     // init syncer ...
} else {
     limiterMu.Unlock()
}
```

It will' be like:

```
var (
    limiterMapOnce = map[string]sync.Once{} // It's empty
)

...

limiterMu.Lock() // anyway lock required to avoid race condition below
if _, ok := limiterMapOnce[p.pipeline]; !ok {
    syncForPipe := sync.Once{}
    limiterMapOnce[p.pipeline] = syncForPipe


    syncForPipe.Do(func() {  // yet this redundant, because next goroutines won't pass if _, ok := limiterMapOnce[p.pipeline]. So struct{}{} instead of sync.Once is enough
           // init syncer
    })
}


```

plugin/action/throttle/redis_limiter.go
```diff
@@ -0,0 +1,188 @@
...
+		if currBucketIdx > 0 {
+			l.totalLimiter.buckets[currBucketIdx] = totalLimiterVal
+		}
+	}
+	l.totalLimiter.mu.Unlock()
+}
+
+// updateKeyLimit reads key limit from redis and updates current limit.
+func (l *redisLimiter) updateKeyLimit(keyPrefixLen int) {
+	l.keyPrefix.WriteString(keySuffix)
```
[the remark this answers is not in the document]
ansakharov: Good point. Moved to config property

plugin/action/throttle/throttle.go
```diff
@@ -103,40 +168,140 @@ func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
```
vitkovskii, line 225: make this code more readable:
1. extract condition contents to separate function, 
2. don't use `else`
3. you don't need sync.Map as it already under mutex
4. use map[string]bool — we don't need to optimize memory here
5. what does `run syncer at single throttle for one pipeline` mean? `run sync only once per pipeline`?

e.g
```
limitersMu.Lock()
if !limiterSyncMap[p.pipeline] {
    limiterSyncMap[p.pipeline] = true
    runSync(p.pipeline)
}
limitersMu.Unlock()
```

plugin/action/throttle/redis_limiter.go
```diff
@@ -0,0 +1,185 @@
...
+	l.syncLocalGlobalLimiters(maxID)
+	l.updateKeyLimit()
+}
+
+func (l *redisLimiter) syncLocalGlobalLimiters(maxID int) {
+	prefix := l.keyPrefix.String()
+
+	for i, ID := range l.bucketIdsForSync {
+		i := i
+		ID := ID
```
vitkovskii, line 121: why you need this?

plugin/action/throttle/redis_limiter.go
```diff
@@ -0,0 +1,185 @@
...
+
+		bucketIdx := l.keyIdxsForSync[i]
+
+		stringID := strconv.Itoa(ID)
+		key := prefix + stringID
```
vitkovskii, line 126: don't append strings, append byte slices

vitkovskii: or buffers

vitkovskii (changes requested):

plugin/action/throttle/throttle.go
ansakharov: About point num 3: it's not sync map, just named sync to show it's purpose. I will rename it to avoid ambiguity

ansakharov: Point 5: Correct, I meant `run sync only once per pipeline`, I renamed with this suggestion.

plugin/action/throttle/redis_limiter.go
ansakharov: It remains from previous optimizations and it stale now. Removed.

plugin/action/throttle/redis_limiter.go
ansakharov: Done, good point with unsafe.

plugin/action/throttle/throttle.go
ansakharov: All done except point 4: map[string]struct{} suits better then map[string]bool

vitkovskii (approved):

vadimalekseev (approved):

PR #202, opened 2022-09-09
change tests output/s3

DSmolonogov:
test TestStartWithMultiBuckets, oft fall on 397 line, i set more friendly time.Sleep. Test began to fall less frequently.

### DIFF
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -819,10 +819,10 @@ func (p *Test) WaitStoppingProcessing(ctx context.Context, limit int) error {
 	for {
 		select {
 		case <-ctx.Done():
-			return fmt.Errorf("time passed out test failed")
+			return ctx.Err()
 		default:
 			//extract backCounter value
-			if int64(limit) == p.eventPool.backCounter.Load() {
+			if int64(limit+p.settings.Capacity) == p.eventPool.backCounter.Load() {
 				return nil
 			}
 			time.Sleep(time.Millisecond * 5)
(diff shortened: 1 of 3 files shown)


vadimalekseev (changes requested):
Stop using reflection and unsafe.

pipeline/pipeline.go
```diff
@@ -808,3 +809,23 @@ func writeErr(w io.Writer, err string) {
...
+type Test struct {
+	*Pipeline
+}
+
+func (p *Test) WaitStoppingProcessing(ctx context.Context, limit int) error {
+	// wait until all events passed pipeline.
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("time passed out test failed")
```
vadimalekseev, line 822: `return ctx.Err()`

plugin/output/s3/s3_test.go
```diff
@@ -407,11 +417,11 @@ func TestStartWithMultiBuckets(t *testing.T) {
 	assert.True(t, size3 > size2)
 }
 
-func newPipeline(t *testing.T, configOutput *Config, objStoreF objStoreFactory) *pipeline.Pipeline {
+func newPipeline(t *testing.T, configOutput *Config, objStoreF objStoreFactory, pipelineCapacity int) *pipeline.Pipeline {
```
vadimalekseev, line 420: 1. Why do we need to set the pipelineCapacity? It is always 4096
2. Can we return pipeline.Test from this function?

ansakharov (changes requested):

PR #245, opened 2022-12-12
add metadata to event

### LINKED ISSUE #244: Feature: plugin add_source_name + minifix add_hostname
1. Store information about hostname in plugin structure
2. Make new plugin add_source_name that add field  with source name information into json

DSmolonogov:
Fixes #244


plugin/action/add_host/add_host.go
```diff
@@ -39,15 +41,18 @@ func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
...
 
 func (p *Plugin) Stop() {
 }
 
 func (p *Plugin) Do(event *pipeline.Event) pipeline.ActionResult {
-	hostname, _ := os.Hostname()
-
-	event.Root.AddFieldNoAlloc(event.Root, p.config.Field).MutateToString(hostname)
-
+	event.Root.AddFieldNoAlloc(event.Root, p.config.Field).MutateToString(p.hostname)
```
vadimalekseev, line 56: Let's make a copy of the `p.hostname`. This is necessary because we can change the bytes inside the node, so the hostname will also change.

I wrote a simple example to demonstrate this:
https://go.dev/play/p/R5AzBtBCPfp

plugin/action/add_source_name/add_source_name.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (p *Plugin) Start(config pipeline.AnyConfig, _ *pipeline.ActionPluginParams) {
+	p.config = config.(*Config)
+}
+
+func (p *Plugin) Stop() {
+}
+
+func (p *Plugin) Do(event *pipeline.Event) pipeline.ActionResult {
+	event.Root.AddFieldNoAlloc(event.Root, p.config.Field).MutateToString(event.SourceName)
```
vadimalekseev, line 46: Do we have to make a copy here? Not sure, but it seems that yes

plugin/action/add_host/add_host.go
vadimalekseev: @vitkovskii, should we handle such cases?

vadimalekseev (approved):

plugin/action/add_source_name/add_source_name.go
```diff
@@ -0,0 +1,48 @@
...
+	plugin.NoMetricsPlugin
+}
+
+// ! config-params
+// ^ config-params
+type Config struct {
+	// > @3@4@5@6
+	// >
+	// > The event field to which put the source name. Must be a string.
+	Field string `json:"field" default:"source_name" required:"true"` // *
```
vitkovskii, line 24: What is "source name", if i'm not familiar with internals of file.d? Is it input plugin name? Or what?

PR #258, opened 2023-01-09
new way to use env variables in config

### LINKED ISSUE #256: Feature: new yaml templating engine
Create a function that will be analogous to applyEnvs(), which will replace environment variables with their value

DSmolonogov:
Fixes #256

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -73,11 +73,6 @@ func NewConfigFromFile(path string) *Config {
 		logger.Fatalf("can't get config values from environments: %s", err.Error())
 	}
 
-	err = applyEnvsV2(json)
-	if err != nil {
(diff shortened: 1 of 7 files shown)


cfg/config.go
```diff
@@ -110,6 +115,43 @@ func applyEnvs(json *simplejson.Json) error {
 	return nil
 }
 
+func applyEnvsV2(json *simplejson.Json) error {
```
vadimalekseev, line 118: Let's make a single function for env variables, not 2 different with different logic:
`applyEnvsV2` applies envs that starts with `$`
`applyEnvs` applies envs that starts with `FILE_D`

vitkovskii (approved):

vadimalekseev (approved):

vano144 (approved):

PR #292, opened 2023-02-05
feat: add ssl and sasl auth for plugin kafka 

akastav:
# Description

Hi! 

This request adds authorization mechanisms to the kafka plugin, as well as support for connection via SSL.

PS: I don't know how to write in GO and there are doubts about the correctness of this compilation, I will be glad if it will be useful. 
The paths to the certificates can probably also be described more elegantly.


plugin/output/kafka/kafka.go
```diff
@@ -188,6 +222,34 @@ func (p *Plugin) Stop() {
...
+		config.Net.SASL.Mechanism = sarama.SASLMechanism(sarama.SASLTypeSCRAMSHA512)
+	}
+
+	// kafka connect via SSL with PEM
+	if p.config.SaslSslEnabled {
+		certs := x509.NewCertPool()
+		pemPath := p.config.SaslPemPath
+		pemData, err := os.ReadFile(pemPath)
+		if err != nil {
+			p.logger.Fatalf("can't load cert: ", err.Error())
```
vadimalekseev, line 242:
```go
p.logger.Fatalf("can't load cert: %s", err.Error())
```

vitkovskii (changes requested):

plugin/output/kafka/kafka.go
```diff
@@ -188,6 +222,34 @@ func (p *Plugin) Stop() {
...
+		}
+		certs.AppendCertsFromPEM(pemData)
```
vadimalekseev, line 244: Can we reuse this logic from elasticsearch output plugin:

https://github.com/ozontech/file.d/blob/master/plugin/output/elasticsearch/elasticsearch.go#L190-L198

It allows to pass path or cert content directly.
It will look like this:
```go
	if p.config.SaslSslEnabled {
		config.Net.TLS.Enable = true

		tlsCfg := tls.NewConfigBuilder()
		if err := tlsCfg.AppendCARoot(p.config.SaslPem); err != nil {
			p.logger.Fatalf("can't load cert: %s", err.Error())
		}
		config.Net.TLS.Config = tlsCfg.Build()
	}
```

vadimalekseev (approved):
LGTM!

vitkovskii (approved):

PR #297, opened 2023-02-08
fix journalctl event reusage in stop function

### LINKED ISSUE #296: Bug: Incorrect reusage of event in journactl input plugin
**Describe the bug**
Incorrect reusage of event in journactl input plugin 

**To Reproduce**
Steps to reproduce the behavior (rare):
1. Two pipelines, one of them is journalctl
2. Too much events from journalctl and other pipeline
3. Need to stop file.d
4. Possible usage of event in stop, but the same event may be in other pipeline

**Expected behavior**
Part of event shouldn't be in offset file.

vano144:
Fixes #296

### DIFF
--- a/plugin/input/journalctl/journalctl.go
+++ b/plugin/input/journalctl/journalctl.go
@@ -120,7 +120,7 @@ func (p *Plugin) Stop() {
 }
 
 func (p *Plugin) Commit(event *pipeline.Event) {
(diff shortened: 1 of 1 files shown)


vadimalekseev (approved):

HeadHunter483 (approved):
LGTM!

PR #312, opened 2023-03-09
311 throttle limiters map cleanup

### LINKED ISSUE #311: Feature: Add limiters map cleanup in throttle plugin
Currently throttle plugin adds limiters to the limiters map depending on the throttle key. The map only grows and limiters are never removed from it. This can cause ineffective memory usage, so the map needs to be cleaned from unused limiters.

HeadHunter483:
Fixes #311

### DIFF
--- a/plugin/action/throttle/limiters_map.go
+++ b/plugin/action/throttle/limiters_map.go
@@ -1,0 +1,221 @@
+package throttle
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/go-redis/redis"
+	prom "github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
+
+	"github.com/ozontech/file.d/pipeline"
+)
+
+const maintenanceInterval = time.Second
+
+// interface with only necessary functions of the original redis.Client
+type redisClient interface {
+	IncrBy(key string, value int64) *redis.IntCmd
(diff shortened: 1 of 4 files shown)


plugin/action/throttle/limiters_map.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		lm.limiterCfg.redisClient = client
+	}
+	return lm
+}
+
+func (l *limitersMap) syncWorker(ctx context.Context, jobCh <-chan limiter, wg *sync.WaitGroup) {
+	for {
+		select {
+		case <-ctx.Done():
```
vadimalekseev, line 104: Deadlock is possible here when we are `wg.Add(1)`, but workers exit here

plugin/action/throttle/limiters_map.go
```diff
@@ -0,0 +1,209 @@
...
+			return
+		case job := <-jobCh:
```
vadimalekseev, line 106: `job` can be `nil` here because you closed the channel before

plugin/action/throttle/limiters_map.go
HeadHunter483: done

plugin/action/throttle/limiters_map.go
HeadHunter483: done

vadimalekseev (approved):

PR #319, opened 2023-03-14
318 throttle fallback redis to in memory

### LINKED ISSUE #318: Feature: Add fallback from redis to in-memory backend in throttle plugin
When redis is chosen as a limiters backend for throttle plugin and the plugin is unable to connect to redis, the whole file.d fatals. Please add a fallback to in-memory backend for that case.

HeadHunter483:
Fixes #318

### DIFF
--- a/plugin/action/throttle/limiters_map.go
+++ b/plugin/action/throttle/limiters_map.go
@@ -2,6 +2,7 @@ package throttle
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


plugin/action/throttle/limiters_map.go
```diff
@@ -114,6 +151,17 @@ func (l *limitersMap) syncWorker(jobCh <-chan limiter, wg *sync.WaitGroup) {
 
 // runSync starts procedure of limiters sync with workers count and interval set in limitersMap.
 func (l *limitersMap) runSync(ctx context.Context, workerCount int, syncInterval time.Duration) {
+	// if redis failed to connect, wait for successful reconnect before starting workers
+	for {
+		l.mu.RLock()
+		if l.isRedisConnected {
```
vano144, line 157: can we use atomic here?

HeadHunter483: done

vano144 (approved):

PR #324, opened 2023-03-25
Antispammer excepts

### LINKED ISSUE #323: Feature: Flexible antispammer settings
Let's support flexible antispammer settings. Currently, all events pass through it. We must begin to separate important logs from the rest and give API to manage this behavior.

The new config will look like this:
```yaml
pipelines:
  k8s:
    settings:
      antispammer:
        threshold: 1000 # moves from settings.antispam_threshold
        pass_policy: 'field_equal' # it may be one of: 'field_equal', 'white_list', 'black_list'

       # policy-specific fields

       # field_equal settings:
       field: 'important'
       value: true

      # white_list/black_list settings:
       field: 'service'
       list: ['payment-api', 's3-server', 'logd']
```

vadimalekseev:
Fixes #323


vano144 (approved):

pipeline/antispam/antispammer.go
```diff
@@ -0,0 +1,188 @@
...
+	Threshold           int
+	UnbanIterations     int
+	Exceptions          []Exception
+
+	Logger            *zap.Logger
+	MetricsController *metric.Ctl
+}
+
+func NewAntispammer(o Options) *Antispammer {
+	if o.Threshold != 0 {
```
HeadHunter483, line 50: I think we should forbid using `o.Threshold` if it is less than zero, or pass to `Antispammer.threshold` zero in that case. It is unlikely that someone on purpose will write negative value to `antispam_threshold` in config, but still this behaviour should be considered.

pipeline/antispam/except.go
```diff
@@ -0,0 +1,155 @@
...
+	Mode   Mode     `json:"mode"`
+	// CaseInsensitive is the truth then Match results in a lowercase search value.
+	// Not available fo the ModeContains because performance issues may arise.
+	CaseInsensitive bool `json:"case_insensitive"`
+	Invert          bool `json:"invert"`
+}
+
+func NewRule(values []string, mode Mode, caseInsensitive bool) Rule {
+	if caseInsensitive && mode == ModeContains {
+		panic(fmt.Errorf("сase insensitive for the 'contains' condtition is not supported"))
```
HeadHunter483, line 48: Typo:
```suggestion
		panic(fmt.Errorf("сase insensitive for the 'contains' condition is not supported"))
```

pipeline/pipeline.go
```diff
@@ -140,6 +142,8 @@ type Settings struct {
 func New(name string, settings *Settings, registry *prometheus.Registry) *Pipeline {
 	metricCtl := metric.New("pipeline_"+name, registry)
 
+	lg := logger.Instance.Named(name)
```
HeadHunter483, line 145: nit: `lg` should be passed to `Pipeline.logger` instead of creating new logger with the same name

vadimalekseev: done

pipeline/antispam/antispammer.go
vadimalekseev: done

pipeline/antispam/except.go
vadimalekseev: done

PR #352, opened 2023-04-12
Non-blocking batcher

### LINKED ISSUE #351: Feature: Speedup batcher
Speedup batcher, it is possible if we release events before waiting for the commit sequence

vadimalekseev:
Fixes #351


pipeline/batch.go
```diff
@@ -150,32 +151,31 @@ func (b *Batcher) work() {
...
-		b.opts.Controller.Commit(e)
+	for i := range events {
+		b.opts.Controller.Commit(events[i], false)
 	}
 
-	b.cond.Broadcast()
-	b.seqMu.Unlock()
-
 	b.freeBatches <- batch
+	b.seqMu.Unlock()
```
vadimalekseev, line 176: I think we need to Unlock first, then Broadcast (?)

pipeline/batch.go
```diff
@@ -150,32 +151,31 @@ func (b *Batcher) work() {
 	}
 }
 
-func (b *Batcher) commitBatch(events []*Event, batch *Batch) []*Event {
-	// we need to release batch first and then commit events
```
vadimalekseev, line 154: I really do not understand this code. This is redundant in my opinion.

pipeline/batch.go
```diff
@@ -227,7 +227,29 @@ func (b *Batcher) getBatch() *Batch {
 func (b *Batcher) Stop() {
 	b.shouldStop.Store(true)
 
-	// todo add scenario without races.
+	b.seqMu.Lock()
```
vadimalekseev, line 230: It fixes the race

vano144 (approved):

PR #368, opened 2023-05-18
Provide metric_skip_status flag to reduce metric cardinality

### LINKED ISSUE #367: Feature: Allow to skip "status" label in the metricHolder
Allow to skip "status" label in the metricHolder to reduce cardinality

vadimalekseev:
Fixes #367

### DIFF
--- a/pipeline/metrics_holder.go
+++ b/pipeline/metrics_holder.go
@@ -98,7 +98,7 @@ func (m *metricsHolder) nextMetricsGen() {
 			continue
 		}
 
-		labels := make([]string, 0, len(metrics.labels))
+		labels := make([]string, 0, len(metrics.labels)+1)
 		if !m.skipStatus {
 			labels = append(labels, "status")
 		}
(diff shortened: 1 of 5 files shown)


pipeline/metrics_holder.go
```diff
@@ -88,11 +92,18 @@ func (c *counter) unregister(registry *prometheus.Registry) {
...
 func (m *metricsHolder) nextMetricsGen() {
 	metricsGen := strconv.Itoa(m.metricsGen % 3) // 2 (for key variance) + 1 (since we must register first) == 3
-	for index, metrics := range m.metrics {
+	for index := range m.metrics {
+		metrics := &m.metrics[index]
 		if metrics.name == "" {
 			continue
 		}
 
+		labels := make([]string, 0, len(metrics.labels))
```
DSmolonogov, line 101: there is reallocation if we add "status" field

pipeline/metrics_holder.go
```diff
@@ -88,11 +92,18 @@ func (c *counter) unregister(registry *prometheus.Registry) {
...
+		if !m.skipStatus {
+			labels = append(labels, "status")
+		}
+		labels = append(labels, metrics.labels...)
```
DSmolonogov, line 105: there is allocation if we add "status" field

PR #376, opened 2023-05-29
Remove map in the rename plugin

### LINKED ISSUE #375: `rename` plugin: Guarantee renaming sequence
Guarantee renaming sequence

vadimalekseev:
Fixes #375


plugin/action/rename/config.go
```diff
@@ -0,0 +1,80 @@
...
+	return "", NotFoundIdx
+}
+
+func (c *Config) Remove(key string) {
+	_, idx := c.Find(key)
+	if idx == NotFoundIdx {
+		return
+	}
+
+	if idx == len(*c) {
```
HeadHunter483, line 36: This condition seems to be always `false` since idx cannot be equal to `len(*c)`

plugin/action/rename/config.go
```diff
@@ -0,0 +1,80 @@
...
+		return errors.New("it is not an object")
+	}
+
+	fields := root.AsFields()
+
+	*c = make([]string, 0, len(fields)*2)
+	for _, field := range fields {
+		key := strings.Clone(field.AsString())
+		value := strings.Clone(field.AsFieldValue().AsString())
+		*c = append(*c, key, value)
```
HeadHunter483, line 76: nit: Maybe it would be better to use `c.Append(key, value)`

HeadHunter483 (approved):

vano144 (approved):

plugin/action/rename/config.go
vadimalekseev: Done :)

Issue #309, opened 2023-03-06
Feature: Run test with -race flag


vadimalekseev:
Let's fix races and run CI test with  the `-race` flag

PR #435, opened 2023-07-24
Fix races, part 2

vadimalekseev:
Related with #309


pipeline/pipeline.go
```diff
@@ -728,6 +728,9 @@ func (p *Pipeline) EnableEventLog() {
 }
 
 func (p *Pipeline) GetEventLogItem(index int) string {
```
vadimalekseev, line 730: Its functions for tests. Move it to another file, with build tag? Or inline this logic to tests

HeadHunter483 (approved):

pipeline/pipeline.go
vano144: maybe just add a comment ?

pipeline/processor.go
```diff
@@ -200,25 +201,25 @@ func (p *processor) doActions(event *Event) (isPassed bool) {
...
 
 			p.finalize(event, false, false)
 			p.actionWatcher.setEventAfter(index, event, eventStatusHold)
-			return false
+			return false, index
 		}
 	}
 
-	return true
+	return true, l - 1
```
vano144, line 222: let's add comment about -1

vano144 (approved):

pipeline/processor.go
vadimalekseev: done

PR #534, opened 2023-10-31
Add move plugin

### LINKED ISSUE #533: Feature: Move plugin
Add new plugin which will be move specified fields to the target field in a certain mode (`allow|block`).
> In `allow` mode, the specified `fields` will be moved;
> in `block` mode, the unspecified `fields` will be moved.

Config:
```
Fields []cfg.FieldSelector
Mode string
Target cfg.FieldSelector
```

kirillov6:
# Description

Added new move plugin.

Fixes #533


plugin/action/move/move.go
```diff
@@ -0,0 +1,229 @@
...
+func factory() (pipeline.AnyPlugin, pipeline.AnyConfig) {
+	return &Plugin{}, &Config{}
+}
+
+func (p *Plugin) Start(config pipeline.AnyConfig, _ *pipeline.ActionPluginParams) {
+	p.config = config.(*Config)
+
+	p.invalidConfig = !p.config.isValid()
+	if p.invalidConfig {
+		return
```
vadimalekseev, line 165: Let's use fatal here

plugin/action/move/move.go
```diff
@@ -0,0 +1,229 @@
...
+			targetNode = targetNode.AddFieldNoAlloc(event.Root, tField).MutateToObject()
+			continue
+		}
+
+		node := targetNode.Dig(tField)
+		if node == nil {
+			node = targetNode.AddFieldNoAlloc(event.Root, tField).MutateToObject()
+			continueDig = false
+		} else if !node.IsObject() {
+			return pipeline.ActionPass
```
vadimalekseev, line 198: Let's overwrite the existing field

plugin/action/move/move.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+	} else {
+		for _, node := range event.Root.AsFields() {
+			value := node.AsFieldValue()
+			if value == targetNode {
+				continue
+			}
+
+			name := node.AsString()
+			if slices.IndexFunc(p.fields, func(f []string) bool { return f[0] == name }) == -1 {
```
vadimalekseev, line 222: I think we should use map here

vadimalekseev (approved):
LGTM!

plugin/action/move/move.go
kirillov6: done

plugin/action/move/move.go
kirillov6: done

plugin/action/move/move.go
kirillov6: done

plugin/action/move/move.go
```diff
@@ -0,0 +1,224 @@
...
+func (p *Plugin) Start(config pipeline.AnyConfig, params *pipeline.ActionPluginParams) {
+	p.config = config.(*Config)
+	if err := p.config.validate(); err != nil {
+		params.Logger.Fatalf("invalid config: %s", err.Error())
+	}
+
+	p.fields = make(map[string][]string)
+	isBlockMode := p.config.Mode == modeBlock
+	for _, fs := range p.config.Fields {
+		// in `block` mode, max field depth is 1
```
vano144, line 170: let's add it to doc

vano144 (approved):

plugin/action/move/move.go
kirillov6: it's already in doc, similar for `target` field
image

PR #537, opened 2023-11-01
Advanced match_fields

### LINKED ISSUE #536: Feature: Add more flexible rules for match_fields
Currently match_fields can only check fields and whether their values are equal to some of the strings. It would be great to have more flexible way to match fields values by prefix, suffix, contains rules as well and combining them with logical operators.

There was a proposal of `do_if` field with more complicated rules than `match_fields` in [this issue](https://github.com/ozontech/file.d/issues/325). And I suggest that it would be interesting to expand it to the tree of nodes, where each node is either logical operator or field operator:

```yaml
pipelines:
  k8s:
    actions:
      - type: 'modify'
        service: 'unknown'
        do_if:
          logical_op: 'or'
          operands:
            - field_op: 'equal'
              field: 'service'
              values:
                - null
                - '' # empty string
            - field_op: 'contains'
              field: 'message'
              values:
                - 'unknown-service'
```

HeadHunter483:
Fixes #536


pipeline/pipeline.go
```diff
@@ -606,7 +608,7 @@ func (p *Pipeline) expandProcs() {
 	}
 
 	for x := 0; x < int(to-from); x++ {
-		proc := p.newProc(p.Procs[from-1].id + x)
```
vano144, line 609: can you add comment about this ? why it is important here and what we are doing

fd/util.go
```diff
@@ -180,6 +185,72 @@ func extractMetrics(actionJSON *simplejson.Json) (string, []string, bool) {
...
+	}
+	result, err = pipeline.NewFieldOpNode(fieldOp, fieldPath, caseSensitive, vals)
+	if err != nil {
+		return nil, fmt.Errorf("failed to init field op: %w", err)
+	}
+
+	return result, nil
+}
+
+func extractDoIfChecker(actionJSON *simplejson.Json) (*pipeline.DoIfChecker, error) {
```
vano144, line 241: nit: test on it will be helpful

HeadHunter483: done

pipeline/pipeline.go
HeadHunter483: done

vadimalekseev (approved):

vano144 (approved):

PR #542, opened 2023-11-08
Regex performance in mask plugin (ignored fields && common match regex)

DmitryRomanov:
# Description

Here are two improvements for mask plugin:

- ignored fields in mask plugin
- compile common match regex (regex_1|regex_2|regex_3) and skip mismatched event before process

### DIFF
--- a/plugin/action/mask/mask.go
+++ b/plugin/action/mask/mask.go
@@ -49,7 +49,8 @@ type Plugin struct {
 	maskBuf []byte
 
 	// common match regex
-	matchRe *regexp.Regexp
+	matchRe       *regexp.Regexp
+	ignoredFields map[string]struct{}
 
 	valueNodes []*insaneJSON.Node
 	logger     *zap.Logger
@@ -69,7 +70,7 @@ type Config struct {
 
 	// > @3@4@5@6
 	// >
-	// > **Experimental feature** for best perfomance. Skips events with mismatched masks.
+	// > **Experimental feature** for best performance. Skips events with mismatched masks.
 	SkipMismatched bool `json:"skip_mismatched" default:"false"` // *
 
 	// > @3@4@5@6
@@ -83,9 +84,8 @@ type Config struct {
 
 	// > @3@4@5@6
 	// >
-	// > List of the ignored event fields.
-	IgnoreFields   []string `json:"ignore_fields"` // *
-	IgnoredFields_ map[string]interface{}
+	// > List of the ignored event fields (including nested fields).
+	IgnoreFields []string `json:"ignore_fields"` // *
 
 	// > @3@4@5@6
 	// >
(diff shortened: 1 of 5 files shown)


plugin/action/mask/mask.go
```diff
@@ -73,6 +81,12 @@ type Config struct {
 	// >
 	MaskAppliedValue string `json:"mask_applied_value"` // *
 
+	// > @3@4@5@6
+	// >
+	// > List of the ignored event fields.
+	IgnoreFields   []string `json:"ignore_fields"` // *
+	IgnoredFields_ map[string]interface{}
```
HeadHunter483, line 88: It would be better to move this map to `Plugin` and rename to `ignoredFields`. Config fields with names `SomeField_` are usually used to store parsed values.

plugin/action/mask/mask.go
```diff
@@ -374,17 +406,19 @@ func (p *Plugin) maskValue(mask *Mask, value, buf []byte) ([]byte, bool) {
 	return value, true
 }
 
-func getValueNodeList(currentNode *insaneJSON.Node, valueNodes []*insaneJSON.Node) []*insaneJSON.Node {
+func getValueNodeList(currentNode *insaneJSON.Node, valueNodes []*insaneJSON.Node, ignoredFields map[string]interface{}) []*insaneJSON.Node {
 	switch {
 	case currentNode.IsField():
-		valueNodes = getValueNodeList(currentNode.AsFieldValue(), valueNodes)
+		if _, ignored := ignoredFields[currentNode.AsString()]; !ignored {
+			valueNodes = getValueNodeList(currentNode.AsFieldValue(), valueNodes, ignoredFields)
```
HeadHunter483, line 413: Please expand a comment on how exactly fields are ignored.

plugin/action/mask/mask.go
```diff
@@ -400,9 +434,13 @@ func (p *Plugin) Do(event *pipeline.Event) pipeline.ActionResult {
 	locApplied := false
 
 	p.valueNodes = p.valueNodes[:0]
-	p.valueNodes = getValueNodeList(root, p.valueNodes)
+	p.valueNodes = getValueNodeList(root, p.valueNodes, p.config.IgnoredFields_)
 	for _, v := range p.valueNodes {
 		value := v.AsBytes()
+		if p.config.SkipMismatched && !p.matchRe.Match(value) {
```
HeadHunter483, line 440: Some masks can have only match rules and no regex but here all masks will be skipped if `SkipMismatched` is true, no matter if they have regex or not. Maybe it would be better to skip only masks with regex in that case.

DmitryRomanov: I added a skip for empty masks and wrote TestWithEmptyRegex

HeadHunter483 (approved):

vano144 (approved):

PR #562, opened 2023-12-20
Fix modify plugin

### LINKED ISSUE #561: Bug: Incorrect modify plugin
**Describe the bug**
The modify plugin doesn't work as described in the documentation.

**To Reproduce**
1. Get config from [doc](https://github.com/ozontech/file.d/blob/master/plugin/action/modify/README.md)
2. Run file.d with this config
3. Instead of replacing the value of the nested field `my_object.field.subfield` - the plugin adds a new field to the root with a name containing "."

**Expected behavior**
The plugin must correctly replace the values of fields whose depth level is greater than 1.

kirillov6:
Fixes #561


plugin/action/modify/modify.go
```diff
@@ -92,7 +100,12 @@ func (p *Plugin) Do(event *pipeline.Event) pipeline.ActionResult {
 			}
 		}
 
-		event.Root.AddFieldNoAlloc(event.Root, field).MutateToBytesCopy(event.Root, p.buf)
+		node := event.Root.Dig(fo.field...)
+		if node == nil {
+			node = event.Root.AddFieldNoAlloc(event.Root, field)
```
HeadHunter483, line 105: Seems like it will add single field `field.subfield` but we want to add nested field `field`->`subfield`. And it would be great to have a test for that case.

kirillov6: done

HeadHunter483 (approved):
LGTM!

vadimalekseev (approved):

vano144 (approved):

PR #568, opened 2024-01-09
Some performance optimizations to the MetricHolder

vadimalekseev:
metricholder takes up a significant amount of CPU on our profile. This PR adds several optimizations. The main optimizations are:

1) Calculating prometheus.MetricVec.WithLabelValues() once, not for every .Inc/.Set/etc
2) Optimization of hash calculation when there is only 1 label
3) Using xxhash for hashing
4) Reduce the amount of memory allocation in the heap
5) Update lastUsage atomic every 10 seconds, not every .Inc/.Set/etc

(sorted by decreasing impact on the benchmark)

old:
```
$ go test -bench='Benchmark(MetricHolder)' -run=XXX -benchmem ./...
goos: darwin
goarch: arm64
pkg: github.com/ozontech/file.d/metric
BenchmarkMetricHolder/l1-8               3806335               299.4 ns/op           240 B/op          6 allocs/op
BenchmarkMetricHolder/l1_l2-8            2969712               404.7 ns/op           288 B/op          6 allocs/op
BenchmarkMetricHolder/l1_l2_l3-8         2462668               487.5 ns/op           336 B/op          6 allocs/op
PASS
ok      github.com/ozontech/file.d/metric       5.006s
```

new:
```
$ go test -bench='Benchmark(MetricHolder)' -run=XXX -benchmem ./...
goos: darwin
goarch: arm64
pkg: github.com/ozontech/file.d/metric
BenchmarkMetricHolder/l1-8              19625188                60.72 ns/op            0 B/op          0 allocs/op
BenchmarkMetricHolder/l1_l2-8           16686858                71.55 ns/op            0 B/op          0 allocs/op
BenchmarkMetricHolder/l1_l2_l3-8        12369921                97.69 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/ozontech/file.d/metric       5.217s
```

prom:
```
$ go test -bench='Benchmark(Prom)' -run=XXX -benchmem ./...
goos: darwin
goarch: arm64
pkg: github.com/ozontech/file.d/metric
BenchmarkPromVec/l1-8            7012634               165.3 ns/op            48 B/op          3 allocs/op
BenchmarkPromVec/l1_l2-8         4787998               250.3 ns/op            96 B/op          3 allocs/op
BenchmarkPromVec/l1_l2_l3-8      3826999               316.3 ns/op           144 B/op          3 allocs/op
PASS
ok      github.com/ozontech/file.d/metric       4.440s
```

TODO: 

- [X] Tests
- [X] Benchmark prometheus.MetricVec
- [X] Benchmark result old/new

### DIFF
--- a/metric/held_metric.go
+++ b/metric/held_metric.go
@@ -50,16 +50,12 @@ func (h *heldMetric[T]) updateUsage() {
 type heldMetricsStore[T prometheus.Metric] struct {
 	mu            sync.RWMutex
 	metricsByHash map[uint64][]*heldMetric[T]
-
-	// used in tests
-	unixNanoFunc func() int64
 }
 
 func newHeldLabelsStore[T prometheus.Metric]() *heldMetricsStore[T] {
 	return &heldMetricsStore[T]{
 		mu:            sync.RWMutex{},
 		metricsByHash: make(map[uint64][]*heldMetric[T]),
-		unixNanoFunc:  xtime.GetInaccurateUnixNano,
 	}
 }
 
@@ -70,6 +66,7 @@ func (h *heldMetricsStore[T]) GetOrCreate(labels []string, createMetric func(...
 	held, ok := h.getHeldLabelsByHash(labels, hash)
 	h.mu.RUnlock()
 	if ok {
+		held.lastUsage.Store(xtime.GetInaccurateUnixNano())
 		return held
 	}
 	// slow path - create new metric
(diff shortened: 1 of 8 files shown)


metric/held_metric.go
```diff
@@ -6,219 +6,125 @@ import (
...
-	// can be replaced for testing
-	isObsoleteFunc func(int64, time.Duration) bool
+	lastUsage atomic.Int64 // unixnano timestamp
+	metric    T
 }
 
-func newHeldLabels(lvs []string) heldLabels {
+func newHeldMetric[T prometheus.Metric](lvs []string, metric T) *heldMetric[T] {
 	labels := make([]string, len(lvs))
+	// copy labels because they are unsafe strings
```
vadimalekseev, line 22: if its unsafe strings, we need to deep copy strings

so we need to replace
`copy(labels, lvs)`
with
```go
labels = slices.Clone(labels)
for i := range labels {
  labels[i] = strings.Clone(labels[i])
}
```
?

vadimalekseev: or why we copy the slice?

PR #619, opened 2024-04-24
dont remove symlink from file provider on broken link

DmitryRomanov:
don't remove symlink from provider in case of broken link (for example in moment of rotation logs in k8s-plugin)


HeadHunter483 (approved):
LGTM!

PR #646, opened 2024-06-25
use franz-go in kafka plugin

DmitryRomanov:
Use [franz-go](https://github.com/twmb/franz-go) for kafka output and input plugins

- add some options for input kafka (heartbeat,session,AWS_MSK_IAM sasl mechanism,auto_commit_interval,balancer_plan,max_concurrent_fetches,fetch_max_bytes,fetch_min_bytes)
- add some options for output kafka (compression,ack)
- remove of using sarama
- fix flaky split_join e2e test


HeadHunter483 (approved):

plugin/input/kafka/client.go
```diff
@@ -0,0 +1,56 @@
...
+		opts = append(opts, kgo.Balancers(kgo.StickyBalancer()))
+	case "cooperative-sticky":
+		opts = append(opts, kgo.Balancers(kgo.CooperativeStickyBalancer()))
+	}
+
+	client, err := kgo.NewClient(opts...)
+	if err != nil {
+		l.Fatalf("can't create kafka client: %s", err.Error())
+	}
+	err = client.Ping(context.TODO())
```
vano144, line 50: nit: context with Tineout ?

cfg/kafka_client.go
```diff
@@ -0,0 +1,113 @@
...
+	ClientCert    string
+	ClientKey     string
+	SslSkipVerify bool
+}
+
+func GetKafkaClientOptions(c KafkaClientConfig, l *zap.SugaredLogger) []kgo.Opt {
+	opts := []kgo.Opt{
+		kgo.SeedBrokers(c.GetBrokers()...),
+		kgo.ClientID(c.GetClientID()),
+		kgo.WithLogger(kzap.New(l.Desugar())),
```
vano144, line 44: nit: log level ?

DmitryRomanov: set by envvar LOG_LEVEL

plugin/input/kafka/client.go
DmitryRomanov: set 1s timeout for ping

```golang
ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
defer cancel()

err = client.Ping(ctx)
if err != nil {
  l.Fatal("can't connect to kafka", zap.Error(err))
}
```

DmitryRomanov: I have decided to use background context in ProduceSync:
Reasons:
- i cannot pick optimal timeout
- we have to wait for result because we send batch
- maybe in [transaction](https://github.com/twmb/franz-go/blob/master/examples/transactions/produce_and_consume/transactional_producer.go) we can set timeout

PR #727, opened 2024-12-23
optimize allocations on pass of offsets to pipeline

DmitryRomanov:
before 

```
BenchmarkLightJsonReadPar-8            1        3706425482 ns/op         391.54 MB/s    135214464 B/op   4192321 allocs/op
```

after

```
BenchmarkLightJsonReadPar-8            1        2704712897 ns/op         536.56 MB/s      226584 B/op       2392 allocs/op
```


HeadHunter483 (approved):

vadimalekseev (approved):
LGTM.

In an internal discussion, it was agreed to optimize other allocations within another PR to remove the regression in version v0.32.0

PR #793, opened 2025-03-02
[#0] - removed duplicate call zap.Error in log

dmpichugin:
# Description

There is already a call to zap.Error in line 522

Fixes # (issue)


vadimalekseev (approved):
Thank you!

HeadHunter483 (approved):

PR #976, opened 2026-05-27
fix: flush accumulated buffer on EOF without trailing newline

### LINKED ISSUE #912: Bug: skips last line without new line
**Describe the bug**

sample of a log file

```json
{"message":"1"}
{"message":"2"} # !!! without \n
```

i am not getting two messages. only first

**To Reproduce**
1. run file.d to process sample log file

**Expected behavior**
in stdout i get only one message

```json
{"message":"1"}
```

but i expect two messages

```json
{"message":"1"}
{"message":"2"}
```

after adding new i get two messages

**Additional context**
Version: v0.71.0
Platform: linux

Vitalymt:
## Problem

When a file doesn't end with a newline character, the last line is silently dropped.

The read loop in `worker.go` accumulates data in `accumBuf` until it finds a `\n`. At EOF, the remaining data is saved to `job.tail` without ever being emitted through `controller.In()`. If the file is never written to again, that data is lost permanently.

**Steps to reproduce:**
1. Create a file with content that doesn't end with `\n`: `echo -n "hello" > /tmp/test.txt`
2. Configure file.d to read this file
3. The "hello" event is never emitted

## Fix

Emit `accumBuf` as a final event before calling `processEOF` when EOF is reached and the buffer is non-empty. Clear both `accumBuf` and `job.tail` afterward to prevent re-emission on file truncation.

**Changes:**
- `plugin/input/file/worker.go`: +8 lines — flush buffer on EOF
- `plugin/input/file/worker_test.go`: updated existing test expectation + added multi-line test case

Fixes #912

### DIFF
--- a/plugin/input/file/worker.go
+++ b/plugin/input/file/worker.go
@@ -202,6 +202,14 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
 		job.tail = append(job.tail[:0], accumBuf...)
 		job.curOffset += readTotal
 
+		// Flush remaining data when EOF is reached and there's no trailing newline.
+		// Without this, the last line of a file that doesn't end with '\n' is silently dropped.
+		if isEOFReached && len(accumBuf) > 0 {
+			job.lastEventSeq = controller.In(sourceID, sourceName, pipeline.NewOffsets(lastOffset+scanned, offsets), accumBuf, isVirgin, metadataInfo)
+			accumBuf = accumBuf[:0]
+			job.tail = job.tail[:0]
+		}
+
 		// check if file was truncated.
 		if isEOFReached {
 			err := w.processEOF(file, job, jobProvider, lastOffset+readTotal)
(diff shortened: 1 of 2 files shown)


plugin/input/file/worker.go
```diff
@@ -202,6 +202,14 @@ func (w *worker) work(controller inputer, jobProvider *jobProvider, readBufferSi
 		job.tail = append(job.tail[:0], accumBuf...)
 		job.curOffset += readTotal
 
+		// Flush remaining data when EOF is reached and there's no trailing newline.
+		// Without this, the last line of a file that doesn't end with '\n' is silently dropped.
+		if isEOFReached && len(accumBuf) > 0 {
```
DmitryRomanov: isEOFReached is checked twice.

It looks cleaner:

```
if isEOFReached {
    // Flush remaining data when EOF is reached and there's no trailing newline.
    if len(accumBuf) > 0 {
        // ...
    }
    err := w.processEOF(...)
}

```

DmitryRomanov (commented):
Please look at the failed tests:
https://github.com/ozontech/file.d/actions/runs/26528924982?pr=976#user-content-tr-BjxUNA-r0s42

plugin/input/file/worker.go
Vitalymt: Good catch, done. Nested the flush inside the single `isEOFReached` block.

PR #1017, opened 2026-09-08
fix(kafka): stop retrying lost consumer membership

### LINKED ISSUE #1016: Bug: Fail fast on UNKNOWN_MEMBER_ID instead of retrying.
We shouldn't retry on UNKNOWN_MEMBER_ID on consume from Kafka. It is a non-retriable protocol error 

If the broker doesn't know the member.id, it means the member has been removed from the group (timeout/rebalance). The broker will never accept this stale ID again. Retrying just creates an infinite loop.

ECD5A:
## Description

When a Kafka consumer group loses membership, franz-go can surface `UNKNOWN_MEMBER_ID` through the fetch path. The previous application-level workaround stopped file.d's only consume loop, which could leave polling and rebalance handling in a bad state.

The franz-go change in [824e34d2](https://github.com/twmb/franz-go/commit/824e34d2), released in v1.21.3, makes classic groups rejoin immediately after `UNKNOWN_MEMBER_ID` or `ILLEGAL_GENERATION` from an offset commit. This PR updates file.d to v1.21.3 and removes the one-way application-level termination path, leaving membership recovery to the Kafka client.

Fixes #1016

## Changes

- update `github.com/twmb/franz-go` from v1.20.7 to v1.21.3;
- update the required module versions and checksums;
- remove the application-level `UNKNOWN_MEMBER_ID` detector and early return.

## Validation

- `go test ./plugin/input/kafka -count=1`
- `go test ./plugin/input/kafka -race -count=1`
- `go vet ./plugin/input/kafka`
- `go mod verify`

The full `go test ./plugin/... -count=1` command was also attempted on Windows. It is currently blocked by unrelated existing failures outside the Kafka plugin, including `syscall.Stat_t` build errors and platform-sensitive file-lock/fixture tests.

A broker-backed before/after reproducer is not included in this repository, so the PR does not claim coverage of every possible `PollRecords` error path. Please confirm whether the upstream change covers the exact scenario from #1016.


DmitryRomanov:
@ECD5A Hello! Maybe this is already fixed in franz-go — [824e34d2](https://github.com/twmb/franz-go/commit/824e34d2), shipped in [v1.21.3](https://github.com/twmb/franz-go/blob/master/CHANGELOG.md#v1213): classic groups now rejoin immediately on UNKNOWN_MEMBER_ID / ILLEGAL_GENERATION from a commit, instead of zombie-consuming until the heartbeat notices.

Can you check the reproduce before and after that fix? If it's already fixed there, we can just bump franz-go

ECD5A:
Thanks for pointing me to 824e34d2. I checked the v1.21.3 change and updated this branch to use franz-go v1.21.3, including the required module updates.

I also removed the application-level `UNKNOWN_MEMBER_ID` early return. In file.d, that return could terminate the only consume loop while `BlockRebalanceOnPoll` was still active.

The focused Kafka tests, race test, vet, and `go mod verify` pass. I do not have a broker-backed before/after reproducer in this repository, so I am not claiming that every `PollRecords` error path is covered. Could you confirm whether 824e34d2 covers the exact scenario from #1016? If so, this dependency-only change should be sufficient.

DmitryRomanov (approved):

timggggggg (approved):

PR #1019, opened 2026-09-20
Fix trim_to delimiter handling

### LINKED ISSUE #1018: Bug: trim_to truncates delimiters and can panic on empty cutsets
**Describe the bug**

The `modify` plugin's `trim_to` filter keeps only the first byte of the matched delimiter when trimming from the right, including `all` mode. Multi-byte delimiters are truncated, and a Unicode delimiter can be cut in the middle of a UTF-8 sequence.

An empty delimiter can cause a slice-bounds panic when the input slice has no spare capacity. With spare capacity, the filter can instead include an extra byte from the backing array.

**To Reproduce**

1. On upstream commit `5379bc2005906fde3aa0a05f6bf574dcd7111404`, run this program from the repository's module:

```go
package main

import (
    "fmt"

    "github.com/ozontech/file.d/cfg/substitution"
    "go.uber.org/zap"
)

func main() {
    for _, tc := range []struct{ delimiter, input string }{
        {"END", "first END second END trailing"},
        {"界", "hello界 trailing"},
        {"", "message"},
    } {
        expr := fmt.Sprintf("${field|trim_to(\"right\",%q)}", tc.delimiter)
        ops, err := substitution.ParseSubstitution(expr, nil, zap.NewNop())
        if err != nil {
            panic(err)
        }
        src := make([]byte, len(tc.input))
        copy(src, tc.input)
        fmt.Printf("%q\n", ops[0].Filters[0].Apply(src, src))
    }
}
```

2. Observe `"first END second E"`, then `"hello\xe7"`, followed by `panic: runtime error: slice bounds out of range [:8] with capacity 7`.

The same right-hand truncation occurs in `all` mode.

**Expected behavior**

Preserve the complete matched delimiter: `"first END second END"` and `"hello界"`. Treat an empty delimiter as leaving the input unchanged, consistently with left trimming.

**Additional context**

Version: upstream master at `5379bc2005906fde3aa0a05f6bf574dcd7111404`.
Platform: macOS arm64, Go 1.27.0.

In `cfg/substitution/trim_to_filter.go`, `bytes.LastIndex` returns the beginning of the final match, but the slice ends at `idx+1`. Using `idx+len(f.cutset)` preserves the whole delimiter and also handles the empty-delimiter case.

I have reproduced the failures in the existing `TestFilterApply` harness and prepared a focused fix with regression tests.

clovn:
# Description

Fix right-side trimming in the `modify` plugin's `trim_to` filter. The slice previously ended one byte after the start of the final delimiter, so `"first END second END trailing"` became `"first END second E"`. It now preserves the complete delimiter, producing `"first END second END"`.

This also prevents partial UTF-8 characters and makes an empty delimiter leave the input unchanged, avoiding a slice-bounds panic or an extra byte from the backing array.

Add eight regression cases covering multi-byte and Unicode delimiters, repeated and missing matches, empty delimiters and input, and left/all modes. No new dependencies.

Fixes #1018

## Validation

Reproduced the failures in the existing test harness before applying the fix. After the fix, these checks passed on macOS arm64 with Go 1.27.0 and golangci-lint 2.13.2:

- `go test ./cfg/substitution -run TestFilterApply -count=1`
- `go test -race ./cfg/... ./plugin/action/modify -count=1`
- `golangci-lint run --config=.golangci.yml --timeout=5m --new-from-rev=upstream/master ./cfg/substitution/... ./plugin/action/modify/...`
- `git diff --check upstream/master...HEAD`

The full CI suite has not been run locally.


kirillov6:
@clovn let's take out the case with the empty `cutset` as erroneous in `trim` and `trim_to` filters.
Add a check on this case in `parseTrimFilter` and `parseTrimToFilter` functions and the corresponding tests in `TestParseSubstitution`.

clovn:
done @kirillov6

kirillov6 (approved):

kirillov6:
LGTM

HeadHunter483:
@clovn please add doc about new behavior in modify plugin https://github.com/ozontech/file.d/blob/master/plugin/action/modify/modify.go#L62

It would be even better to add examples from tests to the doc too.

clovn:
@HeadHunter483 update documentation

HeadHunter483 (approved):
LGTM!

